PR #1, opened 2024-01-31
Use alternate Bezier derivative algorithm

lukem12345:
Hey, cool work on this repo. I like the idea of a line integrals package that could be pulled in when needed.

I was looking at the `derivative` function for Bézier curves, and I noticed that you could take advantage of a simpler trick. You construct a new Bézier curve, one degree lower, whose coefficients are the first differences of those from the original, all multiplied by the original degree. This means you ought to be able to get rid of the constraint on degree.

I did some quick benchmarking with `@btime` and noticed a ~10% performance boost w.r.t. computation time on the first `unit_circle` call to `quadgk` in the tests. I tried out that `unit_circle` example, as well as the code with 1000 vertices instead of 361, although in practice I don’t know if Bézier curves of such high degree are a good idea.

Cheers, Luke

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -14,18 +14,10 @@ function derivative(bz::Meshes.BezierCurve{Dim,T,V}, t) where {Dim,T,V}
     P = bz.controls
     N = degree(bz)
 
-    # Ensure that this implementation is tractible: limited by ability to calculate
-    #   binomial(N, N/2) without overflow. It's possible to extend this range by
-    #   converting N to a BigInt, but this results in always returning BigFloat's.
-    N <= 1028 || error("This algorithm overflows for curves with ⪆1000 control points.")
-
-    # Generator for Bernstein polynomial functions
-    B(i,n) = t -> binomial(Int128(n),i) * t^i * (1-t)^(n-i)
-
-    # Derivative = N Σ_{i=0}^{N-1} sigma(i)
-    #   P indices adjusted for Julia 1-based array indexing
-    sigma(i) = B(i,N-1)(t) .* (P[(i+1)+1] - P[(i)+1])
-    return N .* sum(sigma, 0:N-1)    # ::Vec{Dim,T}
+    # For bz := ∑_{i=0}^{N-1} a_i B_i^N
+    # bz' := N ∑_{i=0}^{N-1} b_i B_i^{N-1},
+    # where b_i := a_{i+1}-a_i.
+    BezierCurve(N*(P[2:end].-P[1:end-1]))(t)
 end
 
 """


mikeingold:
Thanks for checking out the package and for the inputs! I'll test this out as soon as I can.

mikeingold:
I just incorporated this commit locally for testing. I’m not sure just yet what’s going on, but for me it seems to choke on the unit circle example from the README file, which is admittedly very high order. Can you check that one on your side to see if the behavior replicates?

lukem12345:
Hmm, no the time is still similar on my end. I'm testing:
```julia
using DynamicQuantities
using LineIntegrals
using Meshes
using QuadGK
using Unitful
using Test
using BenchmarkTools

####################
# Benchmark Bezier #
####################
unit_circle = BezierCurve(
    [Point(cos(t), sin(t), 0.0) for t in range(0, 2pi, length=361)]
)
f(::Point{Dim,T}) where {Dim,T} = 1.0
@test isapprox(quadgk(f, unit_circle)[1], 2pi; atol=0.15)
@btime quadgk(f, unit_circle)
```
Is this the function to be looking at? If not I may be looking at noise here.

I should note that the `derivative` function in this PR constructs a new Bezier polynomial which when evaluated at `t` returns the derivative there. So whichever interface that repeatedly queries for the derivative would be better off if `derivative` were to return this new Bezier polynomial, and evaluate it as necessary, instead of computing the same coefficients repeatedly.

mikeingold:
This `derivative` function isn't currently utilized by the `integral` or `quadgk` methods, so those tests won't exercise this function. For a minimum working example I'm using the following in a temp environment with only the two packages added:

```julia
using LineIntegrals
using Meshes

unit_circle = BezierCurve(
    [Point(cos(t), sin(t), 0.0) for t in range(0, 2pi, length=361)]
)

derivative(unit_circle, 0.25)
```

I just created a new branch #derivative with this commit applied for testing. I haven't fully waited it out yet, but this MWE seems to completely hang and slowly accrue memory.

```julia
julia> versioninfo()
Julia Version 1.10.0
Commit 3120989f39 (2023-12-25 18:01 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 32 × AMD Ryzen 9 3950X 16-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, znver2)
  Threads: 23 on 32 virtual cores
Environment:
  JULIA_NUM_THREADS = 16
```

lukem12345:
Hmm interesting. I don't understand why this would accrue memory like that, which maybe points to some garbage collecting issue with the Bezier curve function from Meshes?

Either way, I'll push a commit to my branch soon demoing the derivative function as one which returns the derivative of the curve as a curve itself, and then evaluates this single derivative curve for the requested points.

mikeingold:
For the integral method on Bézier curves, I recently figured out that I was getting tons of allocations due to the default method for parametrizing curves. The default is to use DeCasteljau’s algorithm which recursively defines curves of order N-1 until you reach a curve of order 1: so for high-order curves you wind up needing a ton of allocations to handle this.

In this case, though, I’m not really sure what was going on because I wasn’t getting a return. I would be surprised if constructing a single curve and evaluating a parametric point on it would cause an issue like this.

mikeingold:
I haven't absolutely ruled out that there's something unexpected going on in my test environment, so if you're able to test the MWE above and see if the issue duplicates, that's probably a good first step.

mikeingold:
Just a heads-up: I just pushed a fairly major/breaking update bringing the package to `v0.7.0` that expands the scope to include surface and volume integrals on Meshes types. This `derivative` function remains unchanged and still isn’t called by any of the integral solvers, so it shouldn’t directly impact this PR.

PR #11, opened 2024-02-18
Add `SimpleMesh` to roadmap

kylebeggs:
It is great to see numerical packages being built on top of `Meshes.jl`. I already have plans where I can use this in the future, but that really requires being able to integrate over a `SimpleMesh`. I am pretty tied up right now, but I'm trying to stay in the loop on this package and can contribute in a few months when I free up some time. Great work.


mikeingold:
Great idea!  My hope was to build up to more complicated geometries, which I figured in some cases could be integrated by decomposing them into more primitives, at least as a initial implementation.

kylebeggs:
Yep, I agree that is definitely the most sensical approach.

PR #42, opened 2024-08-10
Fix incompatibility with recent Meshes changes

mikeingold:
Implements updates required for compatibility with recent Meshes.j/CRS changes.


mikeingold:
Changes:
- Established compatibility with recent Meshes.jl versions (0.43+) that implemented CRS changes
- Adds direct dependencies on CRS and Unitful for handling of new Meshes.jl type structures
- Added an optional floating-point precision type argument, e.g. `integral(f, geometry, settings, T)`
- Lots of internal restructuring
- Add support for integration of `CylinderSurface`s with `HAdaptiveCubature` algorithm

Longer term outlook (document as TODO actions in GitHub Issues?)
- Evaluate Zygote for AD Jacobians
- Implement generalized GaussLegendre method

Issue #23, opened 2024-03-03
Need improved documentation


mikeingold:
This example code and the resultant benchmarks were run circa v0.9.0 and need to be updated. There's probably no need to include benchmark results here any more, but a demo with a more interesting geometry would be nice.

mikeingold:
## TODO

### README
- [ ] Create a logo and add it on top
- [ ] Write a new package description

### Documenter pages
- [x] Implement a Documenter system
- [ ] Move Example Usage section from README to a Documenter page
- [ ] Move Support Matrix section from README to a Documenter page

mikeingold:
Triangle solver documented in #71

JoshuaLampert:
In my opinion at least the "Example Usage" page needs an update. It would be nice if it is an `@example`, which runs and prints the output. And as you noted above:

> This example code and the resultant benchmarks were run circa v0.9.0 and need to be updated. There's probably no need to include benchmark results here any more, but a demo with a more interesting geometry would be nice.

This is still true.

PR #43, opened 2024-08-11
Update compat bounds for Meshes.jl and set up basic workflows

JoshuaLampert:
Hi @mikeingold! Thanks for this really great package and especially thanks for the recent effort to make MeshIntegrals.jl compatible with the newer versions of Meshes.jl!
I have noticed that MeshIntegrals.jl is not compatible with Meshes.jl <v0.47, which is not a big problem I guess, but this should be reflected by the compat bounds. Also: MeshIntegrals seems to be perfectly compatible with the latest Meshes v0.48. Thus, I think the compat should read "Meshes = 0.47, 0.48".
I also noticed that there are no GitHub actions workflows in this repo. I thus added

- a CompatHelper workflow that automatically creates PRs to update compat bounds, e.g., once a Meshes.jl v0.49 is released CompatHelper will open a PR to update the compat bounds here and thus also allow the new version of Meshes.jl.
- a CI test script that runs the tests on every PR. This also includes code coverage via [codecov](https://about.codecov.io/) and [coveralls](https://coveralls.io/), which can be quite helpful. If you like to upload the coverage reports, you would need to follow the instructions on the linked sites, which is quite easy. If not, I can remove the corresponding lines in the CI script again, of course.
- [dependabot](https://docs.github.com/en/code-security/getting-started/dependabot-quickstart-guide), which allows to get updates for GitHub actions automatically.
- a spell check script, which automatically tests if there are typos in the repository.

Finally, I think, with the recent version bump the syntax for defining a function to be integrated given in the README is outdated. The `p.coords...` gave an error for me. I updated it accordingly.
If you have any questions or suggestions, please feel free to tell me. I put everything in this one PR, but if I should split it up in separate PRs, please also tell me.


mikeingold:
Hello, @JoshuaLampert! Thanks for the PR, and good catch on the compact bounds! It might be a few days before I get a chance to take a deeper look at this but I really appreciate the advice and the contributions.

.github/workflows/CI.yml
mikeingold, line 1: Can you point me toward info on how to setup codecov and coverall?

.github/workflows/CompatHelper.yml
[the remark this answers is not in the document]
JoshuaLampert: Good point. I don't know if it is a problem for CompatHelper.jl, but I removed the `"docs"` in `subdirs` just to be sure.

.github/workflows/CI.yml
JoshuaLampert: Sure!

- codecov: You can visit https://about.codecov.io/ and login with your GitHub account. This will show you a list of your repos. Clicking on "Configure" for this repo will show you three steps to do. You basically only need to copy the `CODECOV_TOKEN` from step 2 and paste it to https://github.com/mikeingold/MeshIntegrals.jl/settings/secrets/actions/new. This token will be used by the script below and coverage reports should be uploaded for each CI run.
-  coveralls: The procedure is pretty much the same. Go to https://coveralls.io/sign-in and sign in with your GitHub account. Turn the switch to "ON" for this repo in the repo list. Clicking on the appearing button "START UPLOAD COVERAGE" again shows you three steps. Like before you need to copy the `COVERALLS_REPO_TOKEN` to the secrets in the GitHub settings of this repo.

JoshuaLampert:
Looks like the tests do not run. Can you double-check the GitHub settings, especially the "Actions" tab under "Code and Automation"?

.github/workflows/CI.yml
mikeingold: Thanks! Done and done

mikeingold:
> Looks like the tests do not run. Can you double-check the GitHub settings, especially the "Actions" tab under "Code and Automation"?

I went in and checked. It looks to me like Actions are allowed to run. What I'm not sure about is whether a proposed CI change will run in the proposed PR, or if only pre-existing CI (none) will run here.

JoshuaLampert:
Usually it should already run in the proposed PR. I am not sure why it does not here, when all settings look fine.

JoshuaLampert:
What is the setting for "Fork pull request workflows from outside collaborators"?

mikeingold:
It was previously defaulted to "require approval for first-time contributors" but I moved it to "...who are new to GitHub" earlier to see if that was the issue. Either way, though, I don't see any options here to approve workflows, just a message that "[CI] has not been set up" with links for me to go add it myself (as was done in this PR).

JoshuaLampert:
Ok, maybe then CI needs to be set up by the owner of a repo to be run in the first PR for which CI it set up?

JoshuaLampert:
We could try merging this PR if you agree and see if it runs then? Or do you have another idea?

JoshuaLampert:
From the [documentation of GitHub Actions about approving workflow runs from public forks](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/approving-workflow-runs-from-public-forks#approving-workflow-runs-on-a-pull-request-from-a-public-fork) it looks like there should be a button "Approve and run" :thinking:

mikeingold:
That's what I expected to see, but it's definitely not there. The question
(possible edge case) of whether or not a PR that implements Actions in the
first place when none pre-exist will still trigger was difficult to search
for an answer on. I tried asking a couple of LLM's and they seemed to think
it might not run, for what little those answers are worth.

In any event, I've read through the changes and don't see any obvious red
flags. I'm inclined to just merge it and then verify it works as expected,
but I'm hoping Julio gets a chance to also give a thumbs up as it would
give me a bit more confidence. This package started off pretty modestly and
I've been running a local "CI" script to run tests on branches in temp
environments, but it's about time to upgrade.

README.md
```diff
@@ -29,15 +33,15 @@ unit_circle_bz = BezierCurve(
 
 # A Real-valued function
 f(x, y, z) = abs(x + y)
-f(p) = f(p.coords...)
+f(p) = f(p.coords.x, p.coords.y, p.coords.z)
```
JoshuaLampert, line 36: Btw, is there a more concise way for this? When I tested `p.coords...` locally, it didn't work anymore. This is maybe also a question for @juliohm.

juliohm: We have two distinct concepts in Meshes.jl now:

1. The `coords` of a point is a complicated struct with important information to locate it in the physical world. It is not a `AbstractVector` nor an iterable. We rarely use `coords` directly in Euclidean geometry.
2. The `to` function is preferred here in the sense that it returns the `Vec` from the origin to the point. It will be a `StaticVector` with support for LinearAlgebra.

I would write the function as

```julia
f(p) = f(to(p)...)
```

JoshuaLampert: Makes sense. Thanks! See [8c5d0a7](https://github.com/mikeingold/MeshIntegrals.jl/pull/43/commits/8c5d0a72fb8ab383b626baffa704a2c8698bdf83).

mikeingold:
On this topic, honestly, the whole of the README probably needs to be
re-written. (open Issue #23). The example code here began in the early
private LineIntegrals.jl days and was only occasionally tweaked/updated.
The choice of BezierCurve was largely to show off the fact that the solver
could handle non-trivial geometries. I'd love to have something more visual
with a more complicated geometry, ideally something that looks like a
real-world problem you could use the package to solve. At that point it
might finally make sense to actually implement a Documenter where there's
more space to show off.

Part of the reason I'd held off on some of this was the uncertainty on
where this package was going to land, i.e. absorbed into Meshes.jl
directly, or transferred into the JuliaGeometry umbrella as a standalone. I
think I'm leaning toward the latter because I hesitate to pull a number of
additional dependencies (QuadGK, HCubature, etc) into Meshes proper.

JoshuaLampert:
I agree that updating the README and creating a proper documentation via Documenter.jl would be nice. But I think this can be done in a subsequent PR. My change in the README in this PR was meant as a quick hotfix.
> I'm hoping Julio gets a chance to also give a thumbs up as it would give me a bit more confidence

Are you fine with merging this PR @juliohm?

mikeingold:
I can see CI action jobs running for the squashed/merged commit.

JoshuaLampert:
Thanks! Things look good: GitHub Actions are [running](https://github.com/mikeingold/MeshIntegrals.jl/actions/runs/10422766340) :partying_face:

PR #44, opened 2024-08-16
Add badges for coverage to README

JoshuaLampert:
Now that we have CI and coverage running, we can add the badges indicating the code coverage and linking to the respective pages to the README.
(This is also meant as a test PR to see if CI runs in a PR now)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 # MeshIntegrals.jl
 
 [![Build Status](https://github.com/mikeingold/MeshIntegrals.jl/actions/workflows/CI.yml/badge.svg?branch=main)](https://github.com/mikeingold/MeshIntegrals.jl/actions/workflows/CI.yml?query=branch%3Amain)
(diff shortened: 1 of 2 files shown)


JoshuaLampert:
Good news: GitHub Action runs.
Bad news: For some reason it only runs the spell checker, but nothing else.

JoshuaLampert:
I see. CI doesn't run because it's [not supposed to run](https://github.com/mikeingold/MeshIntegrals.jl/blob/8e727e8d00869b62226fc4ebf5a033db613fa706/.github/workflows/CI.yml#L17) :sweat_smile:

JoshuaLampert:
With changes to the source it actually runs. So everything looks fine now :+1:. I really seems that GitHub Actions doesn't run if in a PR that creates the workflow and which comes from a fork.

JoshuaLampert:
And codecov comment also works :tada:. So I think we are all set.

PR #45, opened 2024-08-17
Update Meshes compat to v0.49

mikeingold:


JoshuaLampert:
Could you also please put the compat entry of "julia" in Project.toml down as in https://github.com/mikeingold/MeshIntegrals.jl/pull/46/files? Then CompatHelper doesn't want to do that the next time

mikeingold:
Oops, missed that change. Is there a reason why it prefers to have the "julia" entry on the final line? Looks otherwise unchanged.

JoshuaLampert:
I think it's a convention to have it as the final line (?)

JoshuaLampert (approved):

PR #51, opened 2024-08-19
Updated conic integration methods

mikeingold:
## Changes
- New/improved integration methods
  - `Cone`
  - `ConeSurface`
  - `CylinderSurface`
- Support Matrix changes:
  - Enable `CylinderSurface` support for `GaussLegendre`
  - Enable all for `Cone` and `ConeSurface`
  - Remove explicit support for `Frustum` and `FrustumSurface` (#57)
- New tests
  - Custom tests for `Cone` and `ConeSurface`, ...
  - Enable testing for `CylinderSurface` with `GaussLegendre`
- Bumps minimum version compat for Meshes.jl and CoordRefSystems.jl
- Add ColPrac button to README


mikeingold:
Status:
- CylinderSurface passes all tests
- FrustumSurface fails but produces results that match measure to about three decimal places. I suspect this is because measure uses a discretized approximation, i.e. the integral may actually be more accurate.
- ConeSurface fails tests with results that look about 3x too large in the test case.

mikeingold:
It looks like the root issue was that Meshes' parameterization function for `ConeSurface` was producing inaccurate results, seeming to always produce cones whose base has unit radius.

https://github.com/JuliaGeometry/Meshes.jl/issues/1011

Once this upstream bug is sorted out work on this PR will proceed.

mikeingold:
PR pending in Meshes.jl:

https://github.com/JuliaGeometry/Meshes.jl/pull/1020

mikeingold:
The Meshes PR was accepted and merged, awaiting them to push a new patch release (v0.49.6)

PR #63, opened 2024-09-05
Push preview of documentation

JoshuaLampert:
It's nice to see the documentation is evolving!
I saw your question on discourse. This should push a preview of the documentation for each PR. You should be able to visit the preview by clicking on the "Details" link of the "documenter/deploy" CI job below.
I also thought it might be nice to add some more badges to the README and landing page of the docs.


JoshuaLampert:
It also seemed like there were some duplicate docs entries, which made building the docs fail. I fixed that by removing the integration algorithms section. If you prefer another approach, please tell me your preference, @mikeingold.

mikeingold:
Thanks, @JoshuaLampert!

My free time has been a bit fragmented lately, and I made the mistake of not waiting for my last PR to finish Actions before merging (“it’s just a minor update only to the docs” 🤦‍♂️). I was using those doc strings as placeholders for personalized content but hadn’t realized that duplicate `@docs` entries were forbidden. For now, probably best to just un-list the page as you did here.

.github/workflows/DocPreviewCleanup.yml
```diff
@@ -0,0 +1,33 @@
+name: Doc Preview Cleanup
+
+on:
+  pull_request:
+    types: [closed]
```
mikeingold, line 5: I have little experience with GitHub Actions, so pardon any ignorant questions on my part. What does `types: [closed]` indicate? Does it run for all PR's, including drafts?

JoshuaLampert: This new action does not actually perform the preview upload. That is still done by the documenter action from before (now it provides the preview because of the `push_preview=true` keyword in `deploydocs`). This new action only cleans up the GitHub repo by deleting the previews again once they are not needed anymore (i.e. once the corresponding PR is closed). This is only necessary to avoid cluttering the gh-pages branch of the repo where all of the docs previews will be uploaded to.

mikeingold: Ah! That makes sense. Looks like I interpreted this line correctly but missed the context for the action itself

mikeingold:
Thanks again, @JoshuaLampert! You’ve been a huge help maturing this package.

PR #69, opened 2024-09-08
Add TagBot.yml

JoshuaLampert:
I noticed that there is no tag bot used, i.e. you manually create a tag after registering a new version of the package. The tag bot is widely used for Julia packages to automatically create a new tag once the register PR in General is merged. This way consistency between the registry in General and the actual registered versions is guaranteed and you don't need to manually create a new tag. I think it's generally considered good practice to have tag bot running and thus it would be nice to also have it here.
See https://github.com/JuliaRegistries/TagBot.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -9,6 +9,7 @@ on:
       - 'README.md'
       - '.github/workflows/CompatHelper.yml'
       - '.github/workflows/SpellCheck.yml'
+      - '.github/workflows/TagBot.yml'
       - 'docs/**'
   pull_request:
     paths-ignore:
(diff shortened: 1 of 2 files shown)


mikeingold:
Thanks again, @JoshuaLampert!

mikeingold:
@JoshuaLampert What actions are supposed to trigger TagBot to generate a tag and Release? I was thinking it would detect a version bump in the `Project.toml` as the trigger, but their repo README is a little unclear to me in that regard.

JoshuaLampert:
You still need to call the JuliaRegistrator by commenting on the last commit, which bumped the package version as you did before.

PR #70, opened 2024-09-09
README improvements

mikeingold:


mikeingold:
@JoshuaLampert let me know what you think of the updated README in this branch.

README.md
```diff
@@ -11,42 +11,54 @@
...
+Alias functions are provided for convenience. These are simply wrappers for `integral` that first validate that the provided `geometry` has the expected number of parametric/manifold dimensions. Like with `integral` in the examples above, the `algorithm` can also be specified as a third-argument.
+- `lineintegral` for curve-like geometries or polytopes (e.g. `Segment`, `Ray`, `BezierCurve`, `Rope`, etc)
+- `surfaceintegral` for surfaces (e.g. `Disk`, `Sphere`, `CylinderSurface`, etc)
+- `volumeintegral` for (3D) volumes (e.g. `Ball`, `Cone`, `Torus`, etc)
+
+# Demo
 
 ```julia
 using Meshes
 using MeshIntegrals
```
JoshuaLampert, line 49:
```suggestion
using MeshIntegrals
using Unitful
```
You need Unitful here explicitly because you use `u"m"` further down, right?

README.md
```diff
@@ -11,42 +11,54 @@
...
-f(x, y, z) = abs(x + y)
-f(p) = f(to(p)...)
+# Map f(::Point) -> f(x, y, z)
+f(p::Meshes.Point) = f(to(p)...)
 
-integral(f, unit_circle_xy, GaussKronrod())
-    # 0.000170 seconds (5.00 k allocations: 213.531 KiB)
-    # ans == 5.656854249525293 m^2
+# Integrand function in units of Volts/meter
+f(x, y, z) = exp(x^2 + y^2) * u"V/m"
```
JoshuaLampert, line 60:
```suggestion
f(x, y, z) = exp(x^2 + y^2) * u"V/m"
```
Have you tried that this actually works? I had something similar a while back, which didn't work because of the `exp`. Since `x` and `y` have units, I think `exp` throws an error because exponentiating a unit is usually not meaningful.

JoshuaLampert (changes requested):

README.md
mikeingold: Good catch! I’m writing these updates on the go, so haven’t actually tried running them yet. These coordinates are definitely Unitful so won’t work directly inside of `exp`.

JoshuaLampert: You could do something like `f(x, y, z) = exp(u"1/m^2" * (x^2 + y^2)) * u"V/m"`, but I would be interested whether there is another (maybe nicer solution). In my case, `exp` is defined in a function of another package (well, my package...), so it's not so easy to apply the unit within the `exp`.

README.md
mikeingold: Good catch.

README.md
```diff
@@ -11,42 +11,54 @@
...
+volumeintegral(f, geometry)
+```
+Alias functions are provided for convenience. These are simply wrappers for `integral` that first validate that the provided `geometry` has the expected number of parametric/manifold dimensions. Like with `integral` in the examples above, the `algorithm` can also be specified as a third-argument.
+- `lineintegral` for curve-like geometries or polytopes (e.g. `Segment`, `Ray`, `BezierCurve`, `Rope`, etc)
+- `surfaceintegral` for surfaces (e.g. `Disk`, `Sphere`, `CylinderSurface`, etc)
+- `volumeintegral` for (3D) volumes (e.g. `Ball`, `Cone`, `Torus`, etc)
+
+# Demo
 
 ```julia
```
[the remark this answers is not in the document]
JoshuaLampert: I'm sorry. This doesn't make sense. I thought this is within the documentation and not the README.md.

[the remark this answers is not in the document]
mikeingold: I’d like to also significantly expand the Documenter pages with more detailed examples, so that would be a good use in that scenario.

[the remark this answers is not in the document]
mikeingold: Not sure if I should try to tack on those Documenter changes in this PR also or keep this one simple to get it merged sooner.

README.md
JoshuaLampert: I see you simply use `ustrip` there, which of course also works.

README.md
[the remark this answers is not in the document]
JoshuaLampert: I would suggest to separate the two PRs.

README.md
mikeingold: I want to emphasize simplicity and capability, just need to find the right balance. Maybe there’s a cleaner way to modify the wrapper `f(p)` to strip the units there?

mikeingold: I just commit a possible fix for this issue. I’ll hopefully be back on later this evening to run the code and see if it actually works.

README.md
```diff
@@ -11,42 +11,55 @@
...
+# Define a path that approximates a sine-wave on the xy-plane
+mypath = BezierCurve(
+    [Point(t*u"m", sin(t)*u"m", 0.0u"m") for t in range(-pi, pi, length=361)]
 )
 
-# A Real-valued function
-f(x, y, z) = abs(x + y)
-f(p) = f(to(p)...)
+# Map f(::Point) -> f(x, y, z)
+f(p::Meshes.Point) = f(ustrip.(u"m", to(p)...))
```
JoshuaLampert, line 58:
```suggestion
f(p::Meshes.Point) = f(ustrip(to(p))...)
```
I guess, this should work.

README.md
JoshuaLampert: I just ran it and it didn't work. I proposed a solution below, which worked for me.

README.md
JoshuaLampert: An alternative would be `f(p::Meshes.Point) = f(ustrip.(u"m", to(p))...)`, but I prefer the other one because it is simpler.

mikeingold: Fair point. I use the two-argument version out of habit but I guess in this case there’s no need.

mikeingold:
@JoshuaLampert do you have an opinion on bumping the package version within this PR vs separately? I’ve currently got the main branch protected from direct commits so it’s a bit extra overhead to PR a version bump alone.

JoshuaLampert:
It's totally fine to bump the version number in this PR.

README.md
```diff
@@ -11,42 +11,55 @@
...
-ẑ = Vec(0,0,1)
-xy_plane = Plane(origin,ẑ)
-unit_circle_xy = Circle(xy_plane, 1.0)
-
-# Approximate unit_circle_xy with a high-order Bezier curve
-unit_circle_bz = BezierCurve(
-    [Point(cos(t), sin(t), 0.0) for t in range(0,2pi,length=361)]
+# Define a path that approximates a sine-wave on the xy-plane
+mypath = BezierCurve(
+    [Point(t*u"m", sin(t)*u"m", 0.0u"m") for t in range(-, pi, length=361)]
```
JoshuaLampert, line 54: There is something (probably a `pi`?) missing in the first argument of `range`, right?

mikeingold:
On my side it looks like all review items were marked as resolved but I’m still seeing a red flag for a remaining requested changes. Am I missing anything before merging?

README.md
mikeingold: Fixed.

mikeingold:
Am I missing any review items? I’m still seeing a flag for 1 missing suggestion but everything looks resolved on my end.

JoshuaLampert (approved):
Thanks!

JoshuaLampert:
Everything looks fine on my end.

mikeingold:
Great. Thanks!

JoshuaLampert:
Oh, I wanted to enable auto-merge, but missed that CI is not a required status check. I suggest to make passing CI required.

PR #78, opened 2024-09-22
Use TestItems.jl

JoshuaLampert:
As discussed, I reimplemented the testing workflow by using [TestItems.jl](https://github.com/julia-vscode/TestItems.jl) and [TestItemRunner.jl](https://github.com/julia-vscode/TestItemRunner.jl), see also https://www.julia-vscode.org/docs/stable/userguide/testitems/. This makes local development of tests much more convenient and faster by being able to only run certain test items in VScode. I split up the test suite into multiple files because TestItemRunner.jl uses the file structure to group the different test items.


mikeingold:
Does TestItems just automatically parse all of the other files under the tests directory? I'm surprised there's no need to `include` them.

JoshuaLampert:
> Does TestItems just automatically parse all of the other files under the tests directory? I'm surprised there's no need to include them.

Yes, exactly.

mikeingold:
LGTM, thanks!

mikeingold (approved):

PR #79, opened 2024-09-22
Non-breaking internal restructuring

mikeingold:
## Completed
- Major internal re-stucturing and re-organization
  - Moved `IntegrationAlgorithm`s from `integrals.jl` to a new file
  - Consolidated all non-specialized `_integral` workers into `integrals.jl`
  - Moved specialized geometry-specific methods from `integral_*.jl` files into geometry-specific files under the new `/src/specializations/` directory
  - Made `_integral` the first-line worker for all integrals; the only remaining `_integral_Nd` methods are subordinate to `_integral(f, ::GaussKronrod)` methods
  - Culled some obsoleted and redundant method definitions
- Generalized `GaussLegendre` integrals to a single n-dimensional method
- Allow `differential` and `jacobian` to accept argument `ts` as a `Tuple`
- Marginally loosened a few test tolerances (`rtol=1e-6`) to mitigate occasional test failures
- Bump package version from `v0.13.4` to `v0.13.5` and remove sole author credit

**Note on test coverage:** the `Ring`, `Rope`, and `BezierCurve` methods identified as lacking coverage already existed and were only moved to new files in this branch. Ordinarily I would use this opportunity to define tests for them, but the affected methods are going to be obsoleted in an upcoming PR that converts `FP` to a keyword argument.

## Future Work
- Convert `FP` to a keyword argument.
  - Figure out how to ensure clean pass-through of arguments from alias functions, e.g. `lineintegral` -> `integral` -> `_integral`?
- Rename `IntegrationAlgorithm` to `IntegrationRule`, change references from "algorithm", "settings", etc to "rule"
- Add "Why Specialized" lines to each specialization source file: why does this geometry require specialized methods vs the generalized ones?


mikeingold:
@JoshuaLampert Just a heads-up: I was planning for this branch/PR to become a major version bump, but the first (non-breaking) half of the job wound up being so extensive that I think I'm going to split those plans into two PR's. This reorganization is mostly moving methods into different files, combining some redundant wrapper methods, and implementing a generalized Gauss-Legendre method. I'll need to look over everything again, but it should probably be ready for review in the next day or two.

src/integration_algorithms.jl
```diff
@@ -0,0 +1,44 @@
...
+"""
+struct GaussKronrod <: IntegrationAlgorithm
+    kwargs
+    GaussKronrod(; kwargs...) = new(kwargs)
+end
+
+"""
+    GaussLegendre(n)
+
+Numerically integrate using an `n`'th-order Gauss-Legendre quadrature rule. nodes
```
JoshuaLampert, line 21:
```suggestion
Numerically integrate using an `n`'th-order Gauss-Legendre quadrature rule. Nodes
```

src/integration_algorithms.jl
```diff
@@ -0,0 +1,44 @@
...
+time. If the function is know to have some periodic content, then `n` should
+(at a minimum) be greater than the expected number of periods over the geometry,
+e.g. `length(geometry)/lambda`.
+"""
+struct GaussLegendre <: IntegrationAlgorithm
+    n::Int64
+end
+
+"""
+    GaussKronrod(kwargs...)
```
JoshuaLampert, line 35:
```suggestion
    HAdaptiveCubature(kwargs...)
```

test/Project.toml
```diff
@@ -11,7 +11,7 @@ Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 [compat]
 Aqua = "0.8"
 MeshIntegrals = "0.13"
-Meshes = "0.50, 0.51"
+Meshes = "0.51"
```
JoshuaLampert, line 14: Can you comment on why Meshes.jl v0.51 is necessary now? If everything works with Meshes.jl v0.50 I would still like to support it.

test/Project.toml
```diff
@@ -11,7 +11,7 @@ Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 [compat]
 Aqua = "0.8"
 MeshIntegrals = "0.13"
```
JoshuaLampert, line 13:
```suggestion
```
This is not related to this PR, but I think you don't need MeshIntegrals.jl in the Project.toml in test/ (you can also delete it in the [deps] section, but GitHub doesn't allow me to comment there). See also the comment here: https://docs.julialang.org/en/v1/stdlib/Test/#Testing-Our-Package

src/integral.jl
```diff
@@ -76,49 +31,28 @@ function integral(
...
-
-# If algorithm is HAdaptiveCubature, and FP specified, use the generalized n-dim solution
+# with algorithm and T specified
 function integral(
     f::F,
     geometry::G,
-    settings::HAdaptiveCubature,
-    FP::Type{T}
-) where {F<:Function, G<:Meshes.Geometry, T<:AbstractFloat}
+    settings::I,
```
JoshuaLampert, line 38:
```suggestion
    settings::I = HAdaptiveCubature() ,
```
Would it be possible to fuse this method of `integral` with the other one further up by specifying the default for `settings` in the function? Also: You say that the default integration algorithm is `GaussKronrod()` in 1D and `HAdaptiveCubature()` else, but this construct looks like `HAdaptiveCubature` is the default everywhere? Is that correct or can you point me to where the default `GaussKronrod()` is set?

JoshuaLampert (commented):
Nice! I like having more structure in the code. I didn't check everything in detail (especially I assume the copy/paste is fine). I just left some minor comments and questions below.

src/integral.jl
mikeingold: Good catch on the defaults. I’ll need to take a closer look at this.

test/Project.toml
mikeingold: It’s not necessary. In the next major version I’d planned to pull the minimum CRS and Julia versions inline with those in Meshes, but requiring a newer Meshes isn’t actually necessary.

JoshuaLampert: In that case, I would just keep it like it is.

JoshuaLampert: We can also think about another GitHub Action, which tests if all compat bounds are still valid using https://github.com/julia-actions/julia-downgrade-compat. This action downgrades all dependencies to the minimal allowed versions and runs the tests.

src/integral.jl
mikeingold: I just commit an update that adds conditional default integration rules for the simple two-argument method to address this.

test/Project.toml
mikeingold: Done.

test/Project.toml
mikeingold: Versions reverted.

JoshuaLampert: I'm still seeing `Meshes = "0.51"` :thinking:

src/integral.jl
JoshuaLampert: Looks good, but I still don't fully get why we need two methods `integral` (one with two arguments and one with four arguments). Wouldn't it be simpler to have only one method with four arguments with defaults for `settings` and `FP`? I updated my suggestion above accordingly.

src/integral.jl
```diff
@@ -68,57 +23,38 @@ contrast, increasing `FP` to e.g. `BigFloat` will typically increase precision
...
-
-# If algorithm is HAdaptiveCubature, use the generalized n-dim solution
-function integral(
-    f::F,
-    geometry::G,
-    settings::HAdaptiveCubature
-) where {F<:Function, G<:Meshes.Geometry}
-    _integral(f, geometry, settings)
+    N = Meshes.paramdim(geometry)
+    rule = (N == 1) ? GaussKronrod() : HAdaptiveCubature()
```
JoshuaLampert, line 32: One question about the naming: Now it seems like we have three different names for essentially the same thing: `settings`, `algorithm` (in the docstring) and now also `rule`. Is there some meaning behind the different names or could we always just use the same name to avoid confusion?

test/Project.toml
JoshuaLampert: Thanks!

src/integral.jl
mikeingold: We are on the exact same wavelength here. One of the changes I want to roll in with the other breaking `v0.14` changes in an upcoming PR is to get all of this nomenclature aligned. I’m thinking the abstract type should be `IntegrationRule` and “rule” should be used in place of all prior uses of algorithm, setting, etc.

JoshuaLampert: Ok, if this will be aligned in a future PR, I'm fine with that! Thanks for the clarification.

test/Project.toml
mikeingold: Weird, I could have sworn that I commit that edit, but I’m not seeing it either. I’ll double check when I get on.

src/integral.jl
mikeingold: I think this is the way to go but planned to incorporate it into the upcoming breaking change: `FP` becomes a keyword argument, and `rule` can be an optional third argument like this.

JoshuaLampert: Sounds good.

test/Project.toml
mikeingold: Should be fixed now.

JoshuaLampert: Thanks!

JoshuaLampert (approved):

PR #85, opened 2024-09-24
Add downgrade action

JoshuaLampert:
As [discussed](https://github.com/mikeingold/MeshIntegrals.jl/pull/79#discussion_r1771261052) I added a workflow for automatically testing the lower compat bounds. This workflow runs the test with all lower bounds given in the Project.toml. This is considered good practice and helps to keep the compat bounds valid. Running it locally revealed some unsatisfiable requirements, which I fixed by adjusting the lower compat bounds accordingly.


mikeingold:
Interesting that it required those specific minimum versions for Unitful and QuadGK, but they're not overly restrictive and I guess that's why we have the tool now. LGTM.

mikeingold (approved):

PR #86, opened 2024-09-24
Format repo and use formatter workflow

JoshuaLampert:
I ran the JuliaFormatter for the repo with the settings from Meshes.jl, see `.JuliaFormatter.toml`. This results in a lot of changes of course. Sorry for that. Are you open for automatic code formatting to ensure consistency and readability, @mikeingold? Additionally, I added another GitHub Actions workflow to automatically check the formatting for every PR, see also https://github.com/JuliaGeometry/Meshes.jl/pull/965 for more details.


mikeingold:
I like the idea of using something like JuliaFormatter to ensure consistency with a common code style, but I'll probably need to think more about which particular code style to enforce. As-is, I've landed on a particular style in this repo that is pretty self-consistent but kind of esoteric and could use an update. Personally, I think I prefer [SciMLStyle](https://github.com/SciML/SciMLStyle), which seems slightly different from these auto-recommendations.

JoshuaLampert:
Personally, I also prefer the SciMLStyle. This is also the one I use in my repos. I just used the style from Meshes.jl as a first proposal to get the discussion started. I'll quickly reformat. Then we can see how it looks like.

JoshuaLampert:
I'm too lazy to resolve all the conflicts. I'll just close this PR, create a new branch, reformat there and create a new PR.

PR #88, opened 2024-09-25
v0.14-DEV: Remove `derivative` and `unitdirection` from exports

mikeingold:
## Completed
- Removed `derivative` and `unitdirection` from exports. These functions were developed early on when the direction of this package wasn't fully clear, and only `BezierCurve`-specific methods were added. In the long-term, I'd like to implement a more robust system for calculating derivatives and Jacobians, ideally using automatic differentiation. I'm removing these from the public API to facilitate future experimentation but won't remove them from the package entirely.

### DIFF
--- a/src/MeshIntegrals.jl
+++ b/src/MeshIntegrals.jl
@@ -9,7 +9,6 @@ import HCubature
 import QuadGK
 
 include("utils.jl")
-export jacobian, derivative, unitdirection
 
 include("integration_rules.jl")
 export IntegrationRule, GaussKronrod, GaussLegendre, HAdaptiveCubature
--- a/src/MeshIntegrals.jl
+++ b/src/MeshIntegrals.jl
@@ -9,6 +9,7 @@ import HCubature
 import QuadGK
 
 include("utils.jl")
+export jacobian
 
 include("integration_rules.jl")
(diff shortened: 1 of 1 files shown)


JoshuaLampert:
Why is `jacobian` still `export`ed?

mikeingold:
> Why is `jacobian` still `export`ed?

The `derivative` and `unitdirection` (unit vector in same direction as derivative) methods are only defined for Bézier curves and seem mostly like a half-baked idea that we’re never fully implemented. I think my plan was to define these methods for all geometries, but it became obvious to me that plan would be fragile to any upstream changes in how geometries are parameterized. Ideally we could use AD as a general solution for partial derivatives of all geometry types, but I’ve had no luck yet in getting that working out-of-the-box despite some enthusiastic declarations upstream.

For now, I wound up just implementing an inelegant but reasonably effective finite difference approximation within `jacobian`. I was also going to drop `jacobian` but decided that it was at least usable for different geometry types, interesting enough that people might be interested in using it directly, and had a reasonably-stable API. There’s definitely room for improvement, but I figured this one could stick around.

JoshuaLampert (approved):
Ok, thanks for the explanation. If you think `jacobian` is reasonably stable and could be useful in its own, I'm fine exporting it (although I think it's a bit tangential to the goal of the package).

PR #89, opened 2024-09-25
Add code comments justifying use of specialized methods

mikeingold:
## Completed
- Added comment blocks to each `specialized` source file explaining why specialized methods were required for each particular geometry type.

Overall v0.14-DEV roadmap being tracked in #84


JoshuaLampert (approved):
Thanks a lot! That really helps!

PR #90, opened 2024-09-25
Add test with ExplicitImports.jl

JoshuaLampert:
I used https://github.com/ericphanson/ExplicitImports.jl to determine, which symbols are needed and imported them explicitly, which is considered a good practice. I also added a test to ensure only explicit imports and also no stale imports.


JoshuaLampert:
By skimming through the the [documentation of ExplicitImport.jl](https://ericphanson.github.io/ExplicitImports.jl/stable/) again, I saw there are some new tests compared to the last time I checked. I also added them. `check_all_qualified_accesses_are_public` fails, e.g., because `BezierEvalMethod` is not exported from Meshes.jl, but used here. But I think that's fine.

test/aqua.jl
mikeingold, line 1: Should this be added to `aqua.jl` or in a new file?

JoshuaLampert: I also thought about that, but since it is also about "Automated QUality Assurance", I thought it's fine.

src/MeshIntegrals.jl
mikeingold, line 1: Am I correct in thinking that
- function names are only required to be declared here when their use is unqualified in code? e.g. calling only `LinearAlgebra.norm` in code would eliminate the need to add it to the list here?
- `using Example: Example` is basically equivalent to `import Example`?

JoshuaLampert: - Yes, but I would still like to explicitly import `norm` because always writing `LinearAlgebra.norm` looks overly long and complicated, I think.
- Yes, see also a few lines above https://docs.julialang.org/en/v1/manual/modules/#using-and-import-with-specific-identifiers,-and-adding-methods:
  > As we will see in the next section `import .NiceStuff` is equivalent to `using .NiceStuff: NiceStuff`.

mikeingold: I was just using norm as an example. I’m really thinking more toward functions like `Meshes.segment` that I’d probably just qualify since it’s used so infrequently and wouldn’t excessively cloud those lines of code. Not necessary to change anything here, just wanted to make sure my mental model of this requirement was valid. I’ve not read much about explicit imports as a code style, but I’ve made it a habit over the years to try to qualify namespaces where reasonable to reduce the mental burden of keeping track where a used function comes from.

JoshuaLampert: Yes, I think your mental model is valid 🙂 For functions like `Meshes.segment` we can indeed think about qualifying instead of explicitly importing. But as you said, this can be done in another PR.

src/MeshIntegrals.jl
```diff
@@ -1,8 +1,8 @@
 module MeshIntegrals
-using CoordRefSystems
-using LinearAlgebra
-using Meshes
-using Unitful
+using CoordRefSystems: CoordRefSystems, CRS
+using LinearAlgebra: LinearAlgebra, I, norm
+using Meshes: Meshes, Line, Plane, Ray, area, degree, plane, segments, volume, ×, ⋅
```
mikeingold, line 4: Where are `×` and `⋅` products defined on `Meshes.vec` vectors? I had assumed that Meshes was defining methods on `LinearAlgebra.:×` and `LinearAlgebra.:⋅` but I don't see any defined in the codebase. However, when I `using Meshes` in the REPL and check the namespace it does seem like they're exporting the symbols, though the docstrings are just the ones from LinearAlgebra.

JoshuaLampert: Yeah, I was also a bit confused. I think Meshes doesn't need to define methods for `⋅` and `×` because in LinearAlgebra (see e.g. https://github.com/JuliaLang/julia/blob/master/stdlib/LinearAlgebra/src/generic.jl#L922) they are already generic and [`Vec`](https://github.com/JuliaGeometry/Meshes.jl/blob/master/src/vectors.jl#L40) from Meshes.jl inherites `SVector`. So the methods from LinearAlgebra are applicable. This means Meshes.jl basically only reexports them from LinearAlgebra. I guess it doesn't matter whether we import them from LinearAlgebra or from Meshes. Maybe it is nicer to imprt them from LinearAlgebra because this is where they are defined? I let you decide.

mikeingold: That was my thinking too, that they just inherit a method from `SVector`. I'm going to try it out on LinearAlgebra. If everything checks out, then I think we should go with that.

mikeingold:
Wow. Is it normal that the Downgrade Action takes >30 minutes to complete?

mikeingold (approved):

JoshuaLampert:
> Wow. Is it normal that the Downgrade Action takes >30 minutes to complete?

I was also a bit surprised. Seems like more recent package versions made some significant performance improvements (which is also nice to see).

Issue #37, opened 2024-04-07
Improvements to differential/Jacobian calculations


mikeingold:
I consolidated all of the previous differential functions into a single function that uses an `if/else` condition on the number of parametric dimensions

https://github.com/mikeingold/MeshIntegrals.jl/blob/5d92c86d07fa6af97d3df024cdf40fe58a50d16b/src/utils.jl#L62-L81

These operations (norm, norm of cross product, triple product) can be generalized as falling out of an exterior/wedge product using Geometric/Clifford Algebra. It would be nice to generalize this operation into something simple like `norm(foldl(wedge, J))` that doesn't require a conditional.

It looks like DifferentialForms.jl could be used but would require some adaptation to translate between `Form`s and vectors.

mikeingold:
Merging this with AD topic.

The above refers to generalizing the differential element function to arbitrary dimensions. However, the Jacobians this is built on are currently finite-difference approximations. Ideally, these partial differentials could be calculated using an automatic differentiation (AD) package.

mikeingold:
## Status Update
- Changes rolled in with v0.14.0 included merging of the previous `derivative` function into `jacobian`.
- I'm still planning to update `differential` hopefully soon with a generalization to any number of dimensions.
- I have no update on AD functionality. The last I checked, none of the big AD packages I tried were able to achieve AD on Meshes.jl parametric functions out-of-the-box, and defining a bunch of chain rules for them is really something that would need to be done upstream in Meshes.jl proper.

mikeingold:
PR #101 implements a generalization of `differential` to any number of parametric dimensions using geometric algebra (via CliffordNumbers.jl).

The remaining part of this issue is in implementing improved `jacobian` calculations.

mikeingold:
I'm re-splitting this Issue to track the AD topic in a re-opened #35

PR #91, opened 2024-09-26
v0.14-DEV: Change `FP` to a keyword argument

mikeingold:
## Completed
- Refactored to convert floating point precision specification `FP` from an optional argument to a keyword argument.
  - Achieved significant reduction in boilerplate and redundant method definitions.
  - Renamed `GaussKronrod`-specific worker functions to `_integral_gk_($N)d`
  - Converts Meshes and Unitful from `using` to `import`s with explicit namespace references in code
  - Splits differentiation functions into separate source file
  - Completely removes the `unitdirection` function
  - Minor updates to code style/practices


src/integral_aliases.jl
```diff
@@ -16,56 +14,28 @@ Rule types available:
...
-    else
-        error("Performing a line integral on a geometry with $N parametric dimensions not supported.")
-    end
-end
-
 function lineintegral(
         f::F,
         geometry::G,
-        rule::I
+        rule::I;
```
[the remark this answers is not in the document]
JoshuaLampert: Otherwise `lineintegral(f, geometry)` will not work.

JoshuaLampert (commented):
This looks already quite good, but I think we want all `integral` functions in the specializations/ folder to be `_integral`, right? Otherwise the default integration rule will not propagate correctly.
EDIT: It looks like it's fine to use `integral` rather than `_integral` in the specializations. This is because
```julia
julia> foo(a = 1.0) = a
foo (generic function with 2 methods)

julia> foo(a) = a + 1
foo (generic function with 2 methods)

julia> foo()
2.0
```

I would have expected that this gives `1.0`, but ok...good to know :sweat_smile:

mikeingold:
> This looks already quite good, but I think we want all `integral` functions in the specializations/ folder to be `_integral`, right?

This was my original thought, and I tried to do that during the recent PR that consolidated all of the prior paramdim-specific `_integral_Nd` onto `_integral`, but it generated a ton of errors. I'm a little foggy at the moment, but I seem to remember it being a multiple dispatch issue. Not sure if it would still be an issue after this refactor, though.

mikeingold:
> I would have expected that this gives `1.0`, but ok...good to know 😅

That's a bit puzzling to me, too.

JoshuaLampert:
Yes, I just tried replacing `integral` by `_integral` in the specialization and it gave multiple dispatch errors.

mikeingold:
> Yes, I just tried replacing `integral` by `_integral` in the specialization and it gave multiple dispatch errors.

I may take another look at this later today. Maybe I'm just forgetting the nuance, but I think this should be do-able with these new changes.

I think what I was trying to achieve before was just eliminating multi-step pass-throughs, i.e. i.e. `lineintegral(f, g)` -> `integral(f, g)` -> `_integral(f, g)`. If I remember correctly, the issue was that, by skipping `integral`, all of the specialized methods aren't caught. If we define all of those specialized methods as `_integral` instead, then it seems on the surface like this would fix that particular issue.

I do like the idea of having a smaller number of "entry points".

JoshuaLampert:
I also do like to defining specialized methods as `_integral`, but I have the following issue: Let's take `Ring` as an example. Replacing `integral` by `_integral` in the `Ring` specialization gives:

```julia
julia> using Meshes, MeshIntegrals
  1 dependency successfully precompiled in 2 seconds. 115 already precompiled.

julia> geometry = Ring((0.0, 0.0), (1.0, 0.0), (1.0, 1.0), (0.0, 1.0))
Ring
├─ Point(x: 0.0 m, y: 0.0 m)
├─ Point(x: 1.0 m, y: 0.0 m)
├─ Point(x: 1.0 m, y: 1.0 m)
└─ Point(x: 0.0 m, y: 1.0 m)

julia> f(::Point) = 1.0
f (generic function with 1 method)

julia> integral(f, geometry)
ERROR: MethodError: _integral(::typeof(f), ::Ring{𝔼{2}, CoordRefSystems.Cartesian2D{CoordRefSystems.NoDatum, Unitful.Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}}, CircularArrays.CircularVector{Point{𝔼{2}, CoordRefSystems.Cartesian2D{CoordRefSystems.NoDatum, Unitful.Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}}}, Vector{Point{𝔼{2}, CoordRefSystems.Cartesian2D{CoordRefSystems.NoDatum, Unitful.Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}}}}}}, ::GaussKronrod) is ambiguous.

Candidates:
  _integral(f, geometry, rule::GaussKronrod; kwargs...)
    @ MeshIntegrals ~/.julia/dev/MeshIntegrals/src/integral.jl:39
  _integral(f::F, ring::Ring, rule::I; kwargs...) where {F<:Function, I<:IntegrationRule}
    @ MeshIntegrals ~/.julia/dev/MeshIntegrals/src/specializations/Ring.jl:11

Possible fix, define
  _integral(::F, ::Ring, ::GaussKronrod) where F<:Function

Stacktrace:
 [1] integral(f::typeof(f), geometry::Ring{𝔼{…}, CoordRefSystems.Cartesian2D{…}, CircularArrays.CircularVector{…}}, rule::GaussKronrod; kwargs::@Kwargs{})
   @ MeshIntegrals ~/.julia/dev/MeshIntegrals/src/integral.jl:31
 [2] integral(f::typeof(f), geometry::Ring{𝔼{…}, CoordRefSystems.Cartesian2D{…}, CircularArrays.CircularVector{…}}, rule::GaussKronrod)
   @ MeshIntegrals ~/.julia/dev/MeshIntegrals/src/integral.jl:25
 [3] integral(f::typeof(f), geometry::Ring{𝔼{…}, CoordRefSystems.Cartesian2D{…}, CircularArrays.CircularVector{…}})
   @ MeshIntegrals ~/.julia/dev/MeshIntegrals/src/integral.jl:25
 [4] top-level scope
   @ REPL[4]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

This means, which specialization of `_integral` should be called cannot be determined. We have ` _integral(f, geometry, rule::GaussKronrod; kwargs...)` and ` _integral(f::F, ring::Ring, rule::I; kwargs...) `, which both have a specialization in one argument  (`rule` in the first method and `ring` in the second). We want the second to be called, but how should Julia know that?

mikeingold:
Ah, yeah, that’s a harder problem.

mikeingold (commented):
Good catch!

src/differentiation.jl
```diff
@@ -0,0 +1,125 @@
...
+    return map(∂ₙr, 1:length(ts))
+end
+
+"""
+    differential(geometry, ts)
+
+Calculate the differential element (length, area, volume, etc) of the parametric
+function for `geometry` at arguments `ts`.
+"""
+function differential(
```
JoshuaLampert, line 72: Codecov reports that this function is never called. Do we need this?

mikeingold: `differential` or `derivative`? The `differential` function is called in all of the generic `_integral` methods.

The `derivative` function is the one I mentioned before that I pulled from exports. It's not used anywhere else internally, but was postured as the first (currently only) non-finite-diff partial derivative method.

mikeingold: For `derivative` I'm debating a couple of possible options:
1. Just remove it and move on.
2. Add a `@test` to satisfy coverage and continue to let it free-load until the AD path is sorted out.

JoshuaLampert: I'm sorry. I meant `derivative`. I'm on my phone right now. So it's easier to loose track.
We don't need any derivatives to compute integrals along Bezier curves, do we? So would this ever be used for computing integrals in the future (let's say once the AD part is finished)?
If we will need it for the integral computation, could this at least be tested?

mikeingold: For the general (non-specialized) `_integral` solvers, derivatives are used to compute the differential element (i.e. $dl$, $dS$, $dV$) at each evaluated point in the domain. For non-Euclidean manifolds those wind up being functions of position. These are calculated by taking the Jacobian of the geometry's parametric function, i.e. partial derivative `Meshes.Vec`tors with respect to change in each parametric coordinate, and then computing the norm of the exterior products of these `Vec`tors.

Right now, the individual partial derivative `Vec`tors that constitute the Jacobian are calculated via finite-diff inside of the `jacobian` function itself. Ideally, AD would totally replace that functionality inside `jacobian`.

Thinking more about this, I want to go with a **third option:** call this function from the specialized `integral(f, ::BezierCurve)` methods. There's currently a maybe-problematic assumption baked into those methods that the magnitude of $dl$ will be constant like it would be for a line segment, but giving it the full `differential` element treatment to fix this is slightly-overkill.

Do you have a strong opinion on making that a separate PR? I am slightly worried I've scope crept this one since it touched a lot of code anyway.

JoshuaLampert: Ok, that makes sense. Thanks for the detailed explanation. Just for my understanding: Your aim is that ultimately we call `derivative` (for line integrals) within the specialized methods, which are then either computed analytically (e.g. with the current `derivative` for Bezier curves) or using AD? And we would also use AD for every `geometry` without specialization? So we have AD as a general fallback and use analytical derivatives if available?

JoshuaLampert: No need to do anything regarding this in this PR. This is already big enough.

mikeingold: > Just for my understanding: Your aim is that ultimately we call `derivative` (for line integrals) within the specialized methods, which are then either computed analytically (e.g. with the current `derivative` for Bezier curves) or using AD? And we would also use AD for every `geometry` without specialization? So we have AD as a general fallback and use analytical derivatives if available?

Ideally, AD would be capable of calculating `derivative` for all geometries equally. Theoretically, it could be just as good as any analytic solution, and less fragile to implementation changes upstream. Since that capability remains to be seen upstream, for now we'll have to settle on:
- analytic solution for `BezierCurve` defined via `derivative`, and
- the current finite difference fallback/general-solution defined inside `jacobian` for all others

JoshuaLampert: Ok, I would assume that analytical derivatives are more efficient than AD. But I agree that always relying on AD is pretty simple, convenient, and less error-prone. 
But then, do I understand correctly that before implementing AD, you plan to actually use the analytical `derivative` for Bezier curves in the integral computation? Because this analytical derivative becomes obsolete once we use AD, right?

mikeingold: > But then, do I understand correctly that before implementing AD, you plan to actually use the analytical `derivative` for Bezier curves in the integral computation? Because this analytical derivative becomes obsolete once we use AD, right?

Yes. Once this PR merges I'll open another PR that changes all of the `integral(f, ::BezierCurve, rule)` methods to use `derivative` instead of simply domain-correcting with a static `length(curve)` factor. That should improve the accuracy of those calculations, especially for curves with particularly unequal spacing between control points.

Beyond that, I'm cautiously optimistic about using AD to potentially obsolete both the finite-diff and analytic solutions, but it's hard to know how those hopes will pan out until we've got at least a proof of concept AD working on a `Meshes.Geometry` parametric function. I'm planning to track status of this effort here: #37

JoshuaLampert: Alright, sounds like a nice plan. Thanks for your patience explaining me your ideas.

JoshuaLampert (approved):
This looks good to me and can be merged.

mikeingold:
Thanks. I’ll take one last look through the diff to make sure I’ve annotated a list of all the non-trivial changes in the first-post summary, then merge.

mikeingold:
Notes updated at the top. Thanks again for the critical feedback!

PR #92, opened 2024-09-27
Implement `differential` and `jacobian` for Bezier curve integrals

mikeingold:
## Completed
- Merge `derivative` function into `jacobian`
- Use `differential` inside Bezier curve integral methods
- Add a new `@testitem` section for `BezierCurve` with analytic solution
- Add `jacobian` to the Documenter site API page

### DIFF
--- a/src/differentiation.jl
+++ b/src/differentiation.jl
@@ -14,10 +14,10 @@ finite-difference approximation with step size `ε`.
 - `ε`: step size to use for the finite-difference approximation
 """
 function jacobian(
-        geometry,
-        ts;
+        geometry::G,
+        ts::V;
         ε = 1e-6
-)
+) where {G <: Meshes.Geometry, V <: Union{AbstractVector, Tuple}}
     T = eltype(ts)
 
     # Get the partial derivative along the n'th axis via finite difference approximation
@@ -63,6 +63,39 @@ function jacobian(
     return map(∂ₙr, 1:length(ts))
 end
 
+function jacobian(
+    bz::Meshes.BezierCurve,
+    ts::V
+) where {V <: Union{AbstractVector, Tuple}}
+    # Parameter t restricted to domain [0,1] by definition
+    if t < 0 || t > 1
+        throw(DomainError(t, "b(t) is not defined for t outside [0, 1]."))
+    end
+
+    # Aliases
+    P = bz.controls
+    N = Meshes.degree(bz)
+
+    # Ensure that this implementation is tractible: limited by ability to calculate
+    #   binomial(N, N/2) without overflow. It's possible to extend this range by
+    #   converting N to a BigInt, but this results in always returning BigFloat's.
+    N <= 1028 || error("This algorithm overflows for curves with ⪆1000 control points.")
+
+    # Generator for Bernstein polynomial functions
+    B(i, n) = t -> binomial(Int128(n), i) * t^i * (1 - t)^(n - i)
+
+    # Derivative = N Σ_{i=0}^{N-1} sigma(i)
+    #   P indices adjusted for Julia 1-based array indexing
+    sigma(i) = B(i, N - 1)(t) .* (P[(i + 1) + 1] - P[(i) + 1])
+    derivative = N .* sum(sigma, 0:(N - 1))
+
+    return [derivative]
+end
+
+################################################################################
+#                          Differential Elements
+################################################################################
+
 """
     differential(geometry, ts)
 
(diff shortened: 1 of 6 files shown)


JoshuaLampert:
I was just thinking: Could we rename `derivative` to `jacobian`, i.e. add a method to `jacobian`, such that we can reuse `differential` within the Bezier specialization instead of needing `norm(derivative(...))`? Then, of course, the new `jacobian` method (which is now `derivative`) would also get  `ts` instead of `t` (where we would obtain `t` by `t = only(ts)`) and would need to return a one-element vector.

src/specializations/BezierCurve.jl
```diff
@@ -80,9 +80,8 @@ function integral(
         FP::Type{T} = Float64,
         alg::Meshes.BezierEvalMethod = Meshes.Horner()
 ) where {F <: Function, T <: AbstractFloat}
-    len = length(curve)
     point(t) = curve(t, alg)
-    integrand(t) = len * f(point(t[1]))
+    integrand(t) = f(point(only(t))) *  norm(derivative(curve, only(t)))
 
     # HCubature doesn't support functions that output Unitful Quantity types
     # Establish the units that are output by f
```
JoshuaLampert, line 87: Unrelated to this PR, but I just saw: Shouldn't the line below read `testpoint_parametriccoord = zeros(FP, 1)`. I mean, technically it doesn't matter in this case because we apply `only` to it anyway (or before `t[1]`), but the `paramdim` is 1. So the testpoint should also only have length 1, right?

JoshuaLampert: Oh, I see it does matter. `only` doesn't work if the vector has more than one element (even if they are the same).

JoshuaLampert: I fixed that in 25eda748bc00056353d90fb5929f0b63d97ee84d.

mikeingold:
Didn't see you were making edits and just pushed a couple of commits. I have to step away but will check back in a bit to try to sort it out.

JoshuaLampert:
Oh, I'm sorry. I didn't want to disturb your development :grimacing:

mikeingold:
No worries! I'm glad to have the help, just got distracted over in VS Code and was surprised to see help had arrived.

src/specializations/BezierCurve.jl
mikeingold: For some `only` isn't working here as expected, and I'm a bit puzzled as to why.

mikeingold:
I just added a `@testitem` section with new tests that have an analytic solution. I think the core issue we see with test results being non-exact is that these Bezier curves are merely splined/interpolated approximations of the ideal curves. Depending on how the error tolerances stack up, I may need to just lax the `rtol` or `atol` a little.

mikeingold:
> I was just thinking: Could we rename `derivative` to `jacobian`, i.e. add a method to `jacobian`, such that we can reuse `differential` within the Bezier specialization instead of needing `norm(derivative(...))`? Then, of course, the new `jacobian` method (which is now `derivative`) would also get `ts` instead of `t` (where we would obtain `t` by `t = only(ts)`) and would need to return a one-element vector.

This is a really good idea. I think part of the trouble I had conceptually with extending (partial) `derivative` was in trying to generalize it past 1D geometries, i.e. what does the API look like to request the partial derivative w.r.t. the N'th parametric coordinate? Whereas for the 1D case you can just ignore that bit and return a simple `Vec`.

In my mind, then, we've got
```julia
jacobian(geometry::Meshes.Geometry, ts)  # finite-diff fallback
jacobian(curve::Meshes.BezierCurve, ts)  # analytic solution from today's derivative
# etc (as implemented)
```
and can interface them the same in all integral functions.

src/differentiation.jl
```diff
@@ -63,6 +63,39 @@ function jacobian(
     return map(∂ₙr, 1:length(ts))
 end
 
+function jacobian(
+        bz::Meshes.BezierCurve,
+        ts::V
+) where {V <: Union{AbstractVector, Tuple}}
+    # Parameter t restricted to domain [0,1] by definition
```
JoshuaLampert, line 70:
```suggestion
    t = only(ts)
    # Parameter t restricted to domain [0,1] by definition
```

mikeingold:
Pending this latest commit passing CI, this should be good to go.

JoshuaLampert (approved):
Looks good! Thanks @mikeingold!

PR #93, opened 2024-09-28
v0.14.0 Release Candidate

mikeingold:
## Changes
- Removes `-DEV` suffix from version number
- Adds tests to complete code coverage
  - `jacobian(::BezierCurve, ts)` where `t` is outside domain `[0,1]` should throw error
  - `integral(f, geometry)` where `paramdim(geometry)>3` is (currently) broken
- Change integrand function from Bézier test set to remove deprecation warnings.

### DIFF
--- a/test/combinations.jl
+++ b/test/combinations.jl
@@ -11,7 +11,6 @@
         ux = ustrip(p.coords.x)
         (1 / sqrt(1 + cos(ux)^2)) * u"Ω/m"
     end
-    
     f(p::Meshes.Point) = f(ustrip(to(p))...)
     fv(p) = fill(f(p), 3)
 
--- a/test/combinations.jl
+++ b/test/combinations.jl
@@ -37,7 +37,7 @@ end
 @testitem "Meshes.Box" setup=[Setup] begin
     # Test for currently-unsupported >3D differentials
     box4d = Box(Point(zeros(4)...), Point(ones(4)...))
-    @test_throws "Issue on GitHub" jacobian(box4d, fill(0.5, 4))
+    @test_throws "Issue on GitHub" Meshes.differential(box4d, fill(0.5, 4))
 end
 
 @testitem "Meshes.Cone" setup=[Setup] begin
--- a/test/combinations.jl
+++ b/test/combinations.jl
@@ -36,6 +36,12 @@
     @test_throws DomainError jacobian(curve, [1.1])
 end
 
+@testitem "Meshes.Box" setup=[Setup] begin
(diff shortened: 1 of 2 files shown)


test/combinations.jl
```diff
@@ -27,6 +29,15 @@
...
     @test_throws "not supported" surfaceintegral(f, curve)
     @test_throws "not supported" volumeintegral(f, curve)
+
+    # Check Bezier-specific jacobian bounds
+    @test_throws DomainError jacobian(curve, [1.1])
+end
+
+@testitem "Meshes.Box" setup=[Setup] begin
+    # Test for currently-unsupported >3D differentials
+    box4d = Box(Point(zeros(4)...), Point(ones(4))...)
```
JoshuaLampert, line 39:
```suggestion
    box4d = Box(Point(zeros(4)...), Point(ones(4)...))
```

mikeingold:
Thanks for the bugfix!

I'm working out some issues with my local dev environment. Ever since we moved to TestItems I've been getting an opaque error message when trying to run tests from VS Code
> An error occurred attempting to run tests: Error: Pending response rejected since connection got disposed

It's probably just something to do with my Julia environment within Code, not particular to TestItems (I was previously running tests locally but directly from a terminal), but I've had to luck finding solutions online so far.

JoshuaLampert:
Hm ok, I don't know about this error. Did you try to run the tests locally from a terminal via `]test`? If yes, does that work?

JoshuaLampert:
Looks like https://github.com/julia-vscode/julia-vscode/issues/3633?

mikeingold:
> Looks like [julia-vscode/julia-vscode#3633](https://github.com/julia-vscode/julia-vscode/issues/3633)?

Should we not have a `compat` section in the test Project? I’m looking around other major projects and it looks like most only have `deps` specified.

JoshuaLampert:
Yes, indeed we should. Good point. I didn't realize that before,

JoshuaLampert:
Wait...but we have?

JoshuaLampert:
Ahh you mean, we should *not* have one? I think we should have one. So IMO it's fine as it is.

mikeingold:
I’m not sure what the “right” answer is, just noticed that the Issue you linked seemed to point to that as a proximate cause, and then I took a quick look around and noticed other major packages without a `[compat]` section in their `test/Project.toml`.

I’m on mobile right now, so not able to thoroughly research it and I don’t see any contra-indications anywhere about it. If it’s just a bug I can workaround by switching my default Julia to 1.11-rc

PR #95, opened 2024-09-29
Add tests for `Meshes.Quadrangle`

JoshuaLampert:
Seems like integrating `Quadrangle`s simply work. I added tests for that and added them to the support matrix in the docs.


JoshuaLampert:
Looking at the primitives and polytopes in https://github.com/JuliaGeometry/Meshes.jl/tree/168557a714149118a90c570432a873c4047b494a/src/geometries, I think we can also add `Ellipsoid` and `Hexahedron` pretty easily. If I understand correctly, it's more difficult to add general `Ngon`s (apart from `Triangle` and `Quadrangle`), `Polyarea`, `Pyramid`, and `Wedge` because they don't define a parametrization. I think apart from that we cover every primitive and polytope in the support matrix. Correct me if I'm wrong, @mikeingold.

JoshuaLampert:
We could also consider `TransformedGeometry` (should work already for a `geometry`, which is currently supported?) and `MultiGeom`s as sum of the integrals of the single geometries, but that probably doesn't work in every case because the geometries could overlap.

mikeingold:
> Seems like integrating `Quadrangle`s simply work.

That's the beauty of the current general solution methods. Early on with this package I was defining methods for each individual geometry type with hand-derived methods for each individual {geometry, rule} combination. I suspect there's a little bit of a performance penalty for geometries with constant differential elements, e.g. `Segment`, `Box`, etc. When time permits I want to explore an idea I have to conditionally define these as constant for these geometry types instead of re-calculating every iteration.

mikeingold:
On the note of polytopes, I think there’s a little bit of ambiguity over which to treat as surfaces or lines. It seems to me like they’re intended to be defined as essentially Ring’s with a set number of points. However, having simplexes where Triangle is a surface and Tetrahedron is a volume would allow us to integrate any arbitrary geometry by decomposing into those simplexes. That being said, it might currently be an abuse to treat Triangle as a surface like I am.

mikeingold (approved):

mikeingold:
Wait, `Quadrangle` is parameterized? That's surprising.

I guess the fact that both `Triangle` and `Quadrangle` parameterize the surface indicates the intent that they represent surfaces, not mere paths between vertices.

mikeingold:
> Looking at the primitives and polytopes in https://github.com/JuliaGeometry/Meshes.jl/tree/168557a714149118a90c570432a873c4047b494a/src/geometries, I think we can also add `Ellipsoid` and `Hexahedron` pretty easily. If I understand correctly, it's more difficult to add general `Ngon`s (apart from `Triangle` and `Quadrangle`), `Polyarea`, `Pyramid`, and `Wedge` because they don't define a parametrization. I think apart from that we cover every primitive and polytope in the support matrix. Correct me if I'm wrong, @mikeingold.

I would be supportive of tracking support level for all possible geometries in the Support Matrix. I’ve been trying to annotate the ones not currently supported with a link to an Issue so that it communicates some context for why and what the path toward support requires.

In general, if Meshes doesn’t provide a parameterization function, the first consideration would probably be to add one there, if it makes sense to do so. Otherwise, maybe it makes sense to do some sort of decomposition into simplexes or other more primitive geometries? My familiarity with Meshes geometries is somewhat limited beyond those currently in the Matrix.

JoshuaLampert:
> Wait, `Quadrangle` is parameterized? That's surprising.
> 
> I guess the fact that both `Triangle` and `Quadrangle` parameterize the surface indicates the intent that they represent surfaces, not mere paths between vertices.

That's what I assume, e.g. `triangle = Triangle((0.0, 0.0), (1.0, 0.0), (0.5, 1.0))` describes the surface, while `rope Rope((0.0, 0.0), (1.0, 0.0), (0.5, 1.0))` describes the path. This also demonstrated by the fact that `boundary(triangle) == rope` (and of course by the parametric functions). I think the same holds for the other polytopes. In other words, I think your implementation of integrating `Triangle`s seems to me like doing the correct thing :+1:

JoshuaLampert:
> In general, if Meshes doesn’t provide a parameterization function, the first consideration would probably be to add one there, if it makes sense to do so.

Yes, that would be nice, but I don't have the time and the need right now to do that. So I think it's fine that we don't support these geometries for now.

Issue #74, opened 2024-09-19
Floating point argument `FP` not working for `Float32`


mikeingold:
When calling `integral(f, geometry, settings, Float32)`, the intent of this package is to internally compute the integral with 32-bit floating-point precision and return a result with this same precision, i.e. a `Unitful.Quantity` with a `Float32` value.

This issue was discovered when implementing #73. For some reason, these new tests still output results with `Float64` values.

Further debugging will be required. I'm not currently sure where the type promotion is coming from.

PR #96, opened 2024-09-30
Fix bugs in jacobian approximation

mikeingold:
## Completed
- Discovered and fixed a bug that reduced accuracy of some integral calculations.
- Discovered and fixed a bug causing #74.
- Improved performance of `jacobian` function, leading to a performance improvement of about 10% in most integral calculations. 
- Unit tests 
  - Added `combinations.jl` tests for `Box` in (parametric) 1D, 2D, and 3D.
  - Enabled `verbose` for `auto_tests.jl` sets, allowing for direct comparison of runtimes versus `combinations.jl` tests.
  - Removed all `Box` tests from `auto_tests.jl` as now obsoleted.


mikeingold:
Whoa. It looks like these tests uncovered a bug that somehow the auto tests weren't finding. For some reason the Jacobian calculation is using a central finite diff near the interval boundaries. The conditional inside `∂ₙr!` is supposed to switch to left or right finite diff when that close to the domain boundaries to prevent this.

mikeingold:
It looks like I accidentally introduced this bug with commit #8d5f81a in
v0.13.5, where I generalized the `jacobian` function to allow `ts::V` where
`V<:Union{AbstractVector, Tuple}`. The internal finite-diff functions made
use of a re-usable buffer vector that was no longer being re-zero'd between
usages. This should have resulted in inaccurate `differential` results, so
I'm surprised that the impact was apparently so small that tests still
passed (though a few received slightly-increased `atol`/`rtol` values).

TODO (maybe in a new branch/PR):
- Overhaul jacobian's finite-diff code: eliminate re-usable buffer without
new allocations
- Re-run prior tests to determine whether explicit/relaxed tolerances are
still required

mikeingold:
Good news! It looks like the current rewrite of jacobian fixes Issue #74.
It turns out the issue was caused by the default epsilon value being
specified as a Float64. When testing with `FP=BigFloat` this value was
being promoted to BigFloat anyway, but when testing with `FP=Float32` this
caused the parametric coordinates to be promoted to Float64.

src/differentiation.jl
```diff
@@ -18,49 +18,35 @@ function jacobian(
...
+    T = eltype(ts)
+    ε = T(ε)
+
+    # Get the partial derivative along the n'th axis via finite difference
+    #   approximation, where ts is the current parametric position
+    function ∂ₙr(ts, n)
+        # Build left/right parametric coordinates with non-allocating iterators 
+        left =  Iterators.map(it -> it[1] == n ? it[2] - ε : it[2], enumerate(ts))
+        right = Iterators.map(it -> it[1] == n ? it[2] + ε : it[2], enumerate(ts))
+
```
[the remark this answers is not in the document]
mikeingold: @JoshuaLampert do you have any tips for cleaning up these anonymous functions? I was trying to define a 2-tuple anonymous function like `((i, t)) -> i == n ? t + ε : t` but kept getting method errors.

mikeingold:
I never cease to be amazed how drastically even just a few allocations can slow down code. I ran some `@benchmark`s and for a 3D `Box` the mean execution time for `jacobian` drops by an order of magnitude with this code change.

JoshuaLampert:
That's really good new. Thanks a lot @mikeingold. Great work! You say you fixed a bug that lead to reduced accuracy. How big was the loss of accuracy approximately? Can we maybe make some tolerances more strict after this fix?

mikeingold:
> That's really good news. Thanks a lot @mikeingold. Great work! You say you fixed a bug that lead to reduced accuracy. How big was the loss of accuracy approximately? Can we maybe make some tolerances more strict after this fix?

I'm honestly a bit puzzled that this bug didn't cause test failures across the board.

Essentially, for a parametricly-3D geometry, the Jacobian should have been finding tangent-like vectors in the parametric directions `[1, 0, 0]`, `[0, 1, 0]`, `[0, 0, 1]`... but in my mind the bug would have instead been finding tangents in the directions `[1, 0, 0]`, `[1, 1, 0]`, `[1, 1, 1]`. Then the differential element gets calculated via the outer products of all of these. Somehow, it seems like the difference was so small that it didn't even fail most tests with default precision.

mikeingold:
Either way, this is a significantly-improved version of the `jacobian` function. It performs exactly the same work without all the extra abstraction and without all of the extra allocations. On my machine, the old `jacobian` runs in a mean time of around 8 us, while the new one is around `800 ns`. At the aggregate level this seems to have reduced CI time by roughly 10%.

Longer-term I'd like to re-evaluate all of the explicit `rtol` values and rules, e.g. `GaussLegendre(100)` is just a holdover from the auto-tests that seemed to be good enough to universally pass all unit tests. Also, weirdly, the current `Box in 2D` tests seem to take twice as long as the `Box in 3D` tests.

mikeingold:
Did you see the comment I made earlier about anonymous function arguments? I think it got rolled into a resolved review.

src/differentiation.jl
```diff
@@ -18,49 +18,34 @@ function jacobian(
...
-    #   where ts is the current parametric position (εv is a reusable buffer)
-    function ∂ₙr!(εv, ts, n)
+    T = eltype(ts)
+    ε = T(ε)
+
+    # Get the partial derivative along the n'th axis via finite difference
+    #   approximation, where ts is the current parametric position
+    function ∂ₙr(ts, n)
+        # Build left/right parametric coordinates with non-allocating iterators 
+        left = Iterators.map(it -> it[1] == n ? it[2] - ε : it[2], enumerate(ts))
```
JoshuaLampert, line 33:
```suggestion
        left = Iterators.map(it -> ((i, t) = it; i == n ? t - ε : t), enumerate(ts))
```
I saw your question. I'm not sure if there is an optimal solution, but this is another option. I don't see a solution, where you can pass something to the anonymous function, which is not a `Tuple`.

src/differentiation.jl
```diff
@@ -18,49 +18,34 @@ function jacobian(
...
+        right = Iterators.map(it -> it[1] == n ? it[2] + ε : it[2], enumerate(ts))
```
JoshuaLampert, line 34:
```suggestion
        right = Iterators.map(it -> ((i, t) = it; i == n ? t + ε : t), enumerate(ts))
```
See above.

JoshuaLampert (commented):
It's indeed strange that this bug didn't cause more trouble before, but anyway. I made a small suggestion regarding the anonymous function, but it's also not ideal. Otherwise this looks good to me. Thanks again!

test/combinations.jl
```diff
@@ -34,10 +34,100 @@
...
+
+@testitem "Meshes.Box 4D" setup=[Setup] begin
+    box = Box(Point(zeros(4)...), Point(ones(4)...))
+
+    f = p -> one(FP)
+
     # Test for currently-unsupported >3D differentials
-    box4d = Box(Point(zeros(4)...), Point(ones(4)...))
-    @test integral(f -> one(Float64), box4d)≈1.0u"m^4" broken=true
+    @test integral(f, box)≈1.0u"m^4" broken=true
```
JoshuaLampert, line 127: Codecov reports that this doesn't cover the error in `differential`. Before, it did cover it, right? Do you know why it doesn't anymore now?

mikeingold: For a while now I thought this was just a weird Codecov error but I just figured it out. There was an undefined reference on line 124 to `FP` versus the actual floating point type desired, so the test on line 127 would've "successfully" broken for the wrong reason, dropping coverage from the one line in `differential`.

src/differentiation.jl
mikeingold: I finally figured it out and just commit updates to these lines. Looks like the parser interprets the extra parenthesis in `((i, t)) -> ...` as superfluous, and needs to see a trailing comma `((i, t),) -> ...` to understand that this is a tuple being destructured.

mikeingold:
Once this gets merged, I'll get #98 merged and release a v0.14.1.

src/differentiation.jl
JoshuaLampert: Ah, that's nice.

PR #101, opened 2024-10-06
Generalize differential element calculation to n-dims

mikeingold:
## Changes
- Uses geometric algebra to generalize `differential` to handle geometries with arbitrary numbers of parametric dimensions.
- Adds CliffordNumbers.jl as a dependency.
- Converts all prior usages of the generic `error` function into throwing more appropriate/specific error types.
- Implements un-broken tests for 4D `Box` now that this is fixed.
- Minor API change to `jacobian` such that `(geometry, ts) -> ::NTuple{paramdim(geometry), Meshes.Vec}`

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,7 +13,7 @@ QuadGK = "1fd47b50-473d-5c70-9696-f719f8f3bcdc"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [compat]
-CliffordNumbers = "0.1.2"
+CliffordNumbers = "0.1.4"
 CoordRefSystems = "0.12, 0.13, 0.14, 0.15"
 FastGaussQuadrature = "1"
 HCubature = "1.5"
--- a/Project.toml
+++ b/Project.toml
@@ -13,7 +13,7 @@ QuadGK = "1fd47b50-473d-5c70-9696-f719f8f3bcdc"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [compat]
-CliffordNumbers = "^0.1.1"
+CliffordNumbers = "^0.1.2"
 CoordRefSystems = "0.12, 0.13, 0.14, 0.15"
 FastGaussQuadrature = "1"
 HCubature = "1.5"
--- a/Project.toml
+++ b/Project.toml
@@ -3,6 +3,7 @@ uuid = "dadec2fd-bbe0-4da4-9dbe-476c782c8e47"
 version = "0.14.1"
 
 [deps]
+CliffordNumbers = "3998ac73-6bd4-4031-8035-f167dd3ed523"
 CoordRefSystems = "b46f11dc-f210-4604-bfba-323c1ec968cb"
 FastGaussQuadrature = "442a2c76-b920-505d-bb47-c5924d526838"
 HCubature = "19dc6840-f33b-545b-b366-655c7e3ffd49"
@@ -12,6 +13,7 @@ QuadGK = "1fd47b50-473d-5c70-9696-f719f8f3bcdc"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [compat]
+CliffordNumbers = "0.1"
 CoordRefSystems = "0.12, 0.13, 0.14, 0.15"
 FastGaussQuadrature = "1"
 HCubature = "1.5"
--- a/Project.toml
+++ b/Project.toml
@@ -13,7 +13,7 @@ QuadGK = "1fd47b50-473d-5c70-9696-f719f8f3bcdc"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [compat]
-CliffordNumbers = "^0.1.2"
+CliffordNumbers = "0.1.2"
 CoordRefSystems = "0.12, 0.13, 0.14, 0.15"
 FastGaussQuadrature = "1"
 HCubature = "1.5"
--- a/Project.toml
+++ b/Project.toml
@@ -13,7 +13,7 @@ QuadGK = "1fd47b50-473d-5c70-9696-f719f8f3bcdc"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [compat]
-CliffordNumbers = "0.1"
+CliffordNumbers = "^0.1.1"
 CoordRefSystems = "0.12, 0.13, 0.14, 0.15"
 FastGaussQuadrature = "1"
 HCubature = "1.5"
--- a/src/differentiation.jl
+++ b/src/differentiation.jl
@@ -92,7 +92,7 @@ function differential(
         geometry::G,
         ts::V
 ) where {M, CRS, G <: Meshes.Geometry{M, CRS}, V <: Union{AbstractVector, Tuple}}
-    # Calculate the Jacobian, convert each Vec -> (units, KVector)
+    # Calculate the Jacobian, convert Vec -> KVector
     J = jacobian(geometry, ts)
     J_kvecs = Iterators.map(_kvector, J)
 
--- a/src/differentiation.jl
+++ b/src/differentiation.jl
@@ -97,7 +97,7 @@ function differential(
     J_kvecs = Iterators.map(_kvector, J)
 
     # Product of units, exterior-product of KVector's
-    uwedge((u1, kv1), (u2, kv2)) -> (u1 * u2, kv1 ∧ kv2)
+    uwedge((u1, kv1), (u2, kv2)) = (u1 * u2, kv1 ∧ kv2)
     (units, element) = foldl(uwedge, J_kvecs)
 
     return LinearAlgebra.norm(element) * units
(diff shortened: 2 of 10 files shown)


JoshuaLampert:
> This is probably a breaking change since it adds a new dependency.

I don't think that adding a dependency counts as breaking change.

mikeingold:
> > This is probably a breaking change since it adds a new dependency.
> 
> I don't think that adding a dependency counts as breaking change.

I don’t think I’ve seen this documented as a strict requirement. I may be thinking back to a recent heated discussion between the Julia Slack and Discourse about license changes and how some users felt strongly that certain actions constituted a breaking change that should require a more explicit downstream acceptance.

I don’t feel that strongly about it, and I don’t know how many actual users we have that might care, so it’s admittedly probably a bit of an academic question.

mikeingold:
@JoshuaLampert Do you know a way for us to categorize some `@testitem`s or `@tests`s specifically as belonging to some kind of non-CI extended tests? With this PR as-is, Tetrahedron and 4D Box tests are something like 90% of all CI time, so I'm wondering if we can partition them into a separate non-default set of tests.

mikeingold:
Also, thanks again for implementing Downgrade action! It made a good catch here.

JoshuaLampert:
> @JoshuaLampert Do you know a way for us to categorize some @testitems or @testss specifically as belonging to some kind of non-CI extended tests? With this PR as-is, Tetrahedron and 4D Box tests are something like 90% of all CI time, so I'm wondering if we can partition them into a separate non-default set of tests.

Yes, we can give the tests tags and filter all the tests by the tags, see https://www.julia-vscode.org/docs/stable/userguide/testitems/#Tags. Let's do that in a separate PR?

Project.toml
```diff
@@ -12,6 +13,7 @@ QuadGK = "1fd47b50-473d-5c70-9696-f719f8f3bcdc"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [compat]
+CliffordNumbers = "^0.1.2"
```
JoshuaLampert, line 16:
```suggestion
CliffordNumbers = "0.1.2"
```

The caret is the default, cf. https://pkgdocs.julialang.org/v1/compatibility/#Caret-specifiers.

src/integral.jl
```diff
@@ -50,6 +50,8 @@ function _integral(
         return _integral_gk_2d(f, geometry, rule; kwargs...)
     elseif N == 3
         return _integral_gk_3d(f, geometry, rule; kwargs...)
+    else
+        error("Integrating this geometry type with GaussKronrod not supported.")
```
JoshuaLampert, line 54: Could this become a more specific error type, like `MethodError` (I'm not sure if there exists a better error type)?

JoshuaLampert (commented):
That's pretty cool. Thanks. I can't judge the CliffordNumbers.jl implementation because I'm not an expert in geometrical algebra, but since tests pass, I assume everything is fine.

Project.toml
mikeingold: Will this still allow newer patch versions? My reading of the examples on the Pkg.jl docs was that it seemed more restrictive if you specified patch version.

JoshuaLampert: Yes, it will. Both versions (with and without caret) describe the same compat bounds.

src/integral.jl
mikeingold: Based on the docs, I'm not sure `MethodError` is the right choice here. I think `ArgumentError` might be the most appropriate, if we wanted to get away from from generic error types.

mikeingold:
> That's pretty cool. Thanks. I can't judge the CliffordNumbers.jl implementation because I'm not an expert in geometrical algebra, but since tests pass, I assume everything is fine.

The previous `differential` code takes the Jacobian vectors and conditionally uses the `norm` of: (for 1D) the sole partial derivative vector, (for 2D) the cross-product of the two vectors, (for 3D) the vector triple product of the three vectors. It turns out that, in GA, these are just special cases of the more general solution: take the norm of the exterior (wedge) product of all Jacobian vectors.

The CliffordNumbers.jl package is a very lightweight package that implements GA multivectors and their products internally, although they don't currently allow components to be Unitful. So, this code provides a wrapper for converting `Meshes.Vec` to multivectors while additionally keeping tracking of units, then re-combining results with units. Seems to work pretty well as it passes all tests and performance was roughly comparable to before. There seems to have been a small performance penalty, but I think if I change the output type of `jacobian` to `NTuple` this may be even faster than before.

src/integral.jl
JoshuaLampert: I prefer specific error types. `ArgumentError` seems fine 👍

mikeingold: I went through and changed out all instances of `error(msg)` with `throw`n specific errors.

JoshuaLampert: Thanks! Given the similarity between `_error_unsupported_gk` and `_error_unsupported_combination` would it be possible to fuse these two errors into one `_error_unsupported_combination`?

src/differentiation.jl
```diff
@@ -92,16 +92,13 @@ function differential(
...
-    elseif length(J) == 3
-        return abs((J[1] × J[2]) ⋅ J[3])
-    else
-        error("Not implemented yet. Please report this as an Issue on GitHub.")
-    end
+    # Product of units, exterior-product of KVector's
+    uwedge((u1, kv1), (u2, kv2)) = (u1 * u2, kv1 ∧ kv2)
+    (units, element) = foldl(uwedge, J_kvecs)
+
+    return LinearAlgebra.norm(element) * units
```
JoshuaLampert, line 103:
```suggestion
    units = Unitful.unit(J[1][1])
    element = foldl(∧, J_kvecs)

    return LinearAlgebra.norm(element) * units^length(ts)
```
See also my other comment. While experimenting with this a bit, I came up with this alternative to handle the units. It's a bit shorter. I am not sure which version is preferable, but I thought I'll just leave it here. I you prefer the original approach I'm totally fine with that.

JoshuaLampert (commented):
Thanks for your explanation. I now understand what this does. While experimenting with this a bit, I came up with this alternative to handle the units. It's a bit shorter. I am not sure which version is preferable, but I thought I'll just leave it here. I you prefer the original approach I'm totally fine with that.

src/utils.jl
```diff
@@ -10,3 +10,21 @@ end
 
 # Extract the length units used by the CRS of a Point
 _units(pt::Meshes.Point{M, CRS}) where {M, CRS} = first(CoordRefSystems.units(CRS))
+
+# Meshes.Vec -> (units, ::CliffordNumber.KVector)
+function _kvector(v::Meshes.Vec{Dim, T}) where {Dim, T}
+    units = Unitful.unit(T)
+    ucoords = Iterators.map(x -> Unitful.ustrip(units, x), v.coords)
+    kvec = CliffordNumbers.KVector{1, VGA(Dim)}(ucoords...)
+    return (units, kvec)
```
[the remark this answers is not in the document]
mikeingold: I’ll take another look at this later today. Looking at the source for `Meshes.Vec` confirms that the vector components have units that are the same by definition, so you’re right that this can be simplified a bit.

src/integral.jl
mikeingold: I was looking at the cases where the generic GK-specific error was being called and realized that I was throwing errors on the 3D case and also for all >3D cases. I went ahead and eliminated the GK-specific error function and also consolidated the >=3D cases together.

src/utils.jl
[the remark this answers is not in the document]
mikeingold: It turned out that the `_units` function actually already did what we were looking for, just specifically for a `Point`. I abstracted it to work for all geometry types and used that here.

mikeingold:
I added a `@testitem` specifically to check functions under `utils.jl` that should bring us back up to 100% code coverage.

Also, I did some testing locally with a `jacobian` API that returns as `NTuple` instead of `Vector`. There was a huge performance improvement locally, running some integral tests nearly twice as fast on my local machine. Given the huge difference, and that this generalization is a non-trivial feature update, I'd like to go ahead and consolidate these into a v0.15 update.

src/utils.jl
```diff
@@ -8,5 +8,34 @@ function _gausslegendre(T, n)
...
+    return CliffordNumbers.KVector{1, VGA(Dim)}(ucoords...)
+end
+
+# Meshes.Vec -> (units, ::CliffordNumber.KVector)
+function _ukvector(v::Meshes.Vec{Dim, T}) where {Dim, T}
+    units = Unitful.unit(T)
+    ucoords = Iterators.map(x -> Unitful.ustrip(units, x), v.coords)
+    kvec = CliffordNumbers.KVector{1, VGA(Dim)}(ucoords...)
+    return (units, kvec)
+end
```
JoshuaLampert, line 32:
```suggestion
```
Is this needed anywhere?

JoshuaLampert (approved):
LGTM

mikeingold:
> LGTM

Thanks! I'll give this one final review this evening.

mikeingold:
I made two changes:
- Increased minimum version of CliffordNumbers.jl to v0.1.4 based on [a release note from the package owner](https://discourse.julialang.org/t/ann-cliffordnumbers-jl-fast-static-geometric-algebra-with-multivectors-as-number-subtypes/114073/12?u=mike.ingold).
- Removed `@inline` from a utility function; I hear this is now generally not recommended practice.

Issue #67, opened 2024-09-07
Inadequate unit test coverage


mikeingold:
The current unit testing structure was aimed mainly at testing all possible combinations `{f, G, IA}` where `f` is the integral function name/alias, `G` is a `Meshes.Geometry` sub-type, and `IA` is an integration algorithm. In hindsight, the system I wrote for batching these tests was perhaps a little bit too clever for its own good. It also relies on `Meshes.measure` to produce correct answers for a unit-valued integrand.

In the longer term I think a more robust system might define more interesting test geometries with analytic solutions to check against.

As of 7-Sep-2024, we’re currently at about 77% code coverage per the Codecov report. Most of the gaps seem to be related to wrapper/convenience methods, i.e. ones that simply batch out to the core integration methods.

mikeingold:
I just merged #68 beginning an effort to restructure unit tests.

JoshuaLampert:
#73 continues this effort.

JoshuaLampert:
#78 implements a TestItems.jl workflow.

mikeingold:
### Status Update as of v0.14.0

Code coverage is "officially" up to 100% as of the v0.14.0 release. However, this only verifies that all lines of code within this package are exercised during the tests, not that all possible combinations of integral argument types produce correct results.

Work is still ongoing to convert old-style generated `@testset`'s to new-style (`combinations.jl`) `@testitem`'s.

mikeingold:
### Status Update
As of #100 merging, the previous automatic test generation system (`auto-tests.jl`) has been completely removed in favor of a collection of TestItems.jl `@testitem`s (`combinations.jl`).

Remaining work on this Issue:
- Some `@testitem`s still use a unit integrand (I.e. `f(p) = 1.0`) and fairly simple geometry constructions with benchmark integral value provided by `Meshes.measure`. These should be updated with more interesting functions and geometries with known analytics solutions.

mikeingold:
### Status of Unit Tests with Analytic Solutions

- [ ]
- - [ ]
-

mikeingold:
I have an idea for implementing analytical tests for `Triangle` and `Tetrahedron`, but it would require some modification/bypassing of the current `test/combinations.jl` infrastructure:
- Use the appropriately-dimensioned `Box` tests as a starting point.
- Define multiple `Triangle`/`Tetrahedrons` that collectively span the entire `Box` domain.
- Check the sum of these integrals against the correct solution.

PR #105, opened 2024-10-10
Add tests for `ParametrizedCurve`

JoshuaLampert:
I added tests for the new `ParametrizedCurve` in Meshes.jl. I used a curve defined in Cartesian and one in Polar coordinates to test if different crs work as expected. I also added it to the support matrix and bumped the minimum compat version for Meshes.jl in the tests accordingly. Since I defined the curve to describe a 2D sphere for simplicity, this is also used as an analytical solution for the 2D `Sphere` (see #67).


JoshuaLampert:
Hm, downgrade complains because the tests need Meshes.jl v0.51.20 because this is the first version containing `ParametrizedCurve`, but the compat for Meshes.jl in MeshIntegrals.jl itself is `0.50, 0.51`, thus allowing 0.50. But we don't need to bump the lower compat for Meshes.jl in MeshIntegrals.jl because `ParametrizedCurve` is never used in the source. I'm not sure how to deal with this problem.

mikeingold:
Is it possible to make this test item on a separate environment with a specified minimum version?

mikeingold:
Or maybe some kind of conditional statement on the installed version?

JoshuaLampert:
What we could is to define an environment variable in the CI setup, which is, e.g. equal to "part1" or "part2" and split up the tests into two parts. Let's say the new test item is in part2. Then we could run the Downgrade action only for part1 and all the other CI actions run on both "part1" and "part2".

JoshuaLampert:
This way we would also reduce the CI run time.

JoshuaLampert:
> Or maybe some kind of conditional statement on the installed version?

This could maybe also work.

JoshuaLampert:
> What we could is to define an environment variable in the CI setup, which is, e.g. equal to "part1" or "part2" and split up the tests into two parts. Let's say the new test item is in part2. Then we could run the Downgrade action only for part1 and all the other CI actions run on both "part1" and "part2".

Meshes.jl does something similar for Float32 and Float64 tests.

JoshuaLampert:
>> Or maybe some kind of conditional statement on the installed version?

> This could maybe also work.

There is  `pkgversion` for that. This seems like a simple solution. Let's see if tests pass now.

test/combinations.jl
```diff
@@ -703,13 +750,18 @@ end
...
+    function f(p::P) where {P <: Meshes.Point}
+        ur = hypot(p.coords.x, p.coords.y)
+        r = ustrip(u"m", ur)
+        exp(-r^2)
+    end
     fv(p) = fill(f(p), 3)
 
     # Scalar integrand
-    sol = Meshes.measure(sphere)
+    sol = 2π * radius * exp(-radius^2) * u"m"
```
mikeingold, line 764: Nice!

test/combinations.jl
```diff
@@ -516,6 +516,53 @@ end
...
 end
 
+@testitem "ParametrizedCurve" setup=[Setup] begin
+    # ParametrizedCurve has been added in Meshes v0.51.20
+    # If the version is specified as minimal compat bound in the Project.toml, the downgrade test fails
+    if pkgversion(Meshes) >= v"0.51.20"
+        using CoordRefSystems: Polar
+        using LinearAlgebra: norm
+
+        radius = 4.4
```
mikeingold, line 526:
```suggestion
        # Parameterize a circle centered on origin with specified radius
        radius = 4.4
```

mikeingold: I think adding a comment here might be useful to help us remember what this geometry is modeling.

Issue #103, opened 2024-10-10
Stable branch of Documenter site is stale


mikeingold:
For some reason the Documenter site has not kept up with releases since about `v0.13.2`. The `dev` site is current, but the `stable` site has not been updated in some time. It looks like the `stable` site stopped being updated sometime around #70.

JoshuaLampert:
After some more research I found the following:

* https://discourse.julialang.org/t/documenter-jl-does-not-build-on-tags-anymore-after-renaming-master-to-main/57171
From this I conclude that we should avoid using `[skip ci]` in commit messages. Looking at the commits for the releases of v0.13.4 and v0.14.0 we see that no actions were run, i.e. also not the documentation. This is because both corresponding PRs had commits including `[skip ci]`. After squash-merging all commits in these PRs also the final commit in main had `[skip ci]` in the commit message coming from the squashed commits, such that no actions were run.
* https://github.com/JuliaDocs/Documenter.jl/issues/2137 and https://discourse.julialang.org/t/building-docs-with-documenter-jl/63159
Can you please double-check that the `DOCUMENTER_KEY` secret in the repo is set up correctly? This could be the issue because TagBot [uses it](https://github.com/mikeingold/MeshIntegrals.jl/blob/main/.github/workflows/TagBot.yml#L15) and TagBot was introduced in between v0.13.2 and v0.13.3.
* See also https://discourse.julialang.org/t/latest-version-of-docs-not-published-github-actions-tagbot/50634/19 and https://m3g.github.io/JuliaNotes.jl/stable/publish_docs/ and https://github.com/SKopecz/PositiveIntegrators.jl/issues/55
Especially interesting for how to redeploy the missing documentation for the last few releases.

To conclude, my hope is that the `DOCUMENTER_KEY` is not set up (correctly).
If that is the case we should set it up correctly and create a new release to test if it works. Once it works we can try to manually create new tags for the old versions that are missing the docs as explained in one of the links of the third point above.
If, on the other hand, `DOCUMENTER_KEY` is set up correctly, we could still try to manually create a new tag for, say, v0.13.3 to see whether at least this works in order to get the docs for the tags.

mikeingold:
[...]

PR #106, opened 2024-10-11
Testing fix for Documenter issue #103

mikeingold:
I'm taking a swing at fixing #103.

The [deployment docs](https://documenter.juliadocs.org/stable/man/hosting/#Documentation-Versions) and [`deploydocs` docstring](https://documenter.juliadocs.org/stable/lib/public/#Documenter.deploydocs) aren't very descriptive on this point, but from looking around at other packages this `versions` argument seems like it might be needed.

I suspect this won't retroactively include missed versions, though there's [a note on this point](https://documenter.juliadocs.org/stable/man/hosting/#Fixing-broken-release-deployments) that suggests a workaround is to `git tag` old commits with `v0.xx.x-anything` to re-trigger workflows.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -18,4 +18,6 @@ makedocs(
 
 deploydocs(repo = "github.com/mikeingold/MeshIntegrals.jl.git",
     devbranch = "main",
-    push_preview = true)
+    versions = ["stable" => "v^", "v#.#", "dev" => "dev"],
+    push_preview = true
+)


JoshuaLampert:
I suspect that this will not change anything. `versions = ["stable" => "v^", "v#.#", "dev" => "dev"]` is simply [the default](https://github.com/JuliaDocs/Documenter.jl/blob/a0257f939c2be20c7a68ecc02c4ada8210f29495/src/deploydocs.jl#L203). It is also not necessary in my other repos. What is also a bit annoying is the fact that this is not easy to debug. We always need to merge a PR into main in order to get some feedback whether the attempt worked or didn't.

JoshuaLampert:
I see that for v0.13.2 there are these `create mode ...` lines in the logs, see [here](https://github.com/mikeingold/MeshIntegrals.jl/actions/runs/10753335382/job/29822657294#step:6:64), while for the v0.13.3 they are missing, see [here](https://github.com/mikeingold/MeshIntegrals.jl/actions/runs/10839814447/job/30080919434#step:6:81). This could give us a hint, where things might go wrong.

JoshuaLampert:
There is also no `[new tag]` for v0.13.3 before, see [here](https://github.com/mikeingold/MeshIntegrals.jl/actions/runs/10839814447/job/30080919434#step:6:81).

JoshuaLampert:
Ah, [this one](https://github.com/mikeingold/MeshIntegrals.jl/actions/runs/10839814447/job/30080919434#step:6:20).

JoshuaLampert:
Getting closer... Looks like we get into not getting into [this](https://github.com/JuliaDocs/Documenter.jl/blob/a0257f939c2be20c7a68ecc02c4ada8210f29495/src/deployconfig.jl#L331) `if` branch, but into [this](https://github.com/JuliaDocs/Documenter.jl/blob/a0257f939c2be20c7a68ecc02c4ada8210f29495/src/deployconfig.jl#L347) one, which should not be the case.

mikeingold:
> I suspect that this will not change anything. `versions = ["stable" => "v^", "v#.#", "dev" => "dev"]` is simply [the default](https://github.com/JuliaDocs/Documenter.jl/blob/a0257f939c2be20c7a68ecc02c4ada8210f29495/src/deploydocs.jl#L203).

Damn. Looks like you're correct.

> What is also a bit annoying is the fact that this is not easy to debug. We always need to merge a PR into main in order to get some feedback whether the attempt worked or didn't.

Exactly my feelings, too. Troubleshooting Actions themselves is very frustrating when the existing version of the Action is the one that gets run.

JoshuaLampert:
I think I found the issue. You seemed to have registered a new version in another PR than the one, where you bumped the version number. Namely here: https://github.com/mikeingold/MeshIntegrals.jl/commit/c5ee55711590961f1c28e3ea276d638699c860cb#comments. I think this lead to `GITHUB_REF` being set wrongly and thus Documenter.jl using the wrong if branch as outlined above.

JoshuaLampert:
However, I still don't know why it didn't get fixed with the newer versions starting from v0.13.4.

mikeingold:
Weird. So I bumped the package patch version number in one commit but waited until after the next commit was added to trigger registration, and that somehow broke Documenter moving forward?

JoshuaLampert:
Maybe, I am not sure TBH.

JoshuaLampert:
But I _do_ think, it's a problem with `GITHUB_REF` because we really should go into the `build_type === :release` branch in order to have the correct `subfolder`, where the docs are deployed, set in [this line](https://github.com/JuliaDocs/Documenter.jl/blob/a0257f939c2be20c7a68ecc02c4ada8210f29495/src/deployconfig.jl#L346).

mikeingold:
Yet another reason I'm a fan of Julia and functional programming: fewer worries about magical background state. 😆

PR #109, opened 2024-10-14
Add an extended tag for time-intensive unit tests

mikeingold:
## Changes
- Establishes an `:extended` `@testitem` tag for time-intensive tests that can forgo CI (still available for local testing in VS Code)
  - Tagged `Box in 4D` and `Tetrahedron`
  - 

This comes with the downside that integration of these two geometry types won’t be checked automatically with CI, and a code coverage “reduction”, but I think the actual practical impact is minor. For any changes that would specifically impact these two geometries, we’ll have to make sure to run the tests locally. Slightly inconvenient for a niche case but allows for faster iterations in development of everything else.

### DIFF
--- a/test/combinations.jl
+++ b/test/combinations.jl
@@ -168,7 +168,7 @@ end
     @test volumeintegral(f, box) ≈ sol
 end
 
-@testitem "Meshes.Box 4D" setup=[Setup] begin
+@testitem "Meshes.Box 4D" tags=[:extended] setup=[Setup] begin
     a = π
     box = Box(Point(0, 0, 0, 0), Point(a, a, a, a))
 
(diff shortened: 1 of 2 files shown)


mikeingold:
We're down to <10 minute CI times just by pushing two geometry types into `:extended` test sets. 😃

JoshuaLampert (changes requested):
Could we put the test for a wrong number of parametric coordinates into another `@testitem` to keep at least that one covered? It currently is in the 4D Box test (lines 199-200), but could also live somewhere else.

mikeingold:
Good catch! I just moved it from 4D to 2D

JoshuaLampert (approved):
Thanks!

PR #110, opened 2024-10-16
Add more analytic tests

mikeingold:
## Changes
- Continues to make progress on #67 
  - [x] `Ball` (2D)
  - [x] `Ball` (3D)
  - [x] `Circle`
  - [x] `Disk`
  - [x] `Sphere` (3D)
- Add `LinearAlgebra` to standard test snippet and convert usages from `hypot` to `norm`

### DIFF
--- a/test/combinations.jl
+++ b/test/combinations.jl
@@ -36,24 +36,20 @@ end
     using SpecialFunctions: erf
     using LinearAlgebra: norm
 
-    center = Point(0, 0, 0)
+    center = Point(1, 2, 3)
     radius = 2.8u"m"
     ball = Ball(center, radius)
 
     function f(p::P) where {P <: Meshes.Point}
-        ur = norm(to(p))
+        ur = norm(to(p) - to(center))
         r = ustrip(u"m", ur)
         exp(-r^2)
-        # 1 / r
-        # 1.0
     end
     fv(p) = fill(f(p), 3)
 
     # Scalar integrand
     r = ustrip(u"m", radius)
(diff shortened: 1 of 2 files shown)


mikeingold:
@JoshuaLampert can you look at the `@testitem` section for `Meshes.Ball 3D`? For some reason, the analytic solutions that I'm calculating aren't matching the integral results for any non-unit integrand function. Fingers crossed that I'm just tired and this isn't an actual bug.

JoshuaLampert:
I think there was a mistake in how you compute the radius inside the integration function (you were missing the z component). I changed it to use `norm(to(p))`, which should be dimension agnostic. However, this didn't fix the problem. Your analytical solutions look correct to me. Surprisingly tests pass locally after I changed the center to be (0, 0, 0) 😳. I'll check if there is an error in the parametrization in Meshes.jl because this is really strange. The center should not have any influence on the volume.

JoshuaLampert:
Now I see, you used something similar already for `Disk`, but use `hypot` instead of `norm`. While this of course doesn't make a difference, I find `norm` a bit more intuitive. I think unifying this to one of these, would be nice for consistency and understandability. I slightly tend to use `norm` everywhere, but if you prefer `hypot` I would also be fine.

mikeingold:
I’m very happy to hear that the root cause here was simply my lack of sleep and not something deeper.

I’ll take a look at the other tests and see where it makes sense to use `norm` vs `hypot`.

Thanks!

JoshuaLampert:
I'm sorry. Yesterday was busy for me. But this looks good to me anyway. Thanks!

mikeingold:
No worries. I figured you were busy, and I’d already gotten a partial review anyway.

PR #114, opened 2024-10-21
Use newly-supported KVector Unitful extension to simplify differential

mikeingold:
## Status

Need to rebase after completion of #122.

## Changes
- Increases minimum compat version of CliffordNumbers.jl dependency to `v0.1.9`
- Uses CliffordNumbers.jl's [new Unitful extension](https://github.com/brainandforce/CliffordNumbers.jl/issues/26), converting `Meshes.Vec` to `Unitful.Quantity{KVector}` which now support unit-aware wedge products
- Add a new utility method `_units(::Meshes.Vec)`

PR #118, opened 2024-10-23
Explicitly import Meshes.Geometry

mikeingold:
## Changes
- Explicitly import `Meshes.Geometry`. This is a frequently-used reference with a fairly obvious namespace source, and being able to reference it implicitly will reduce function signature line lengths.
- Update method signatures, eliminating unnecessary `where` clauses. I'm told that there's no functional/performance difference between a function argument being `f(x::SuperType)` vs `f(x::ST) where {ST <: SuperType}`.

### DIFF
--- a/src/MeshIntegrals.jl
+++ b/src/MeshIntegrals.jl
@@ -1,11 +1,11 @@
 module MeshIntegrals
 using CliffordNumbers: CliffordNumbers, VGA, ∧
 using CoordRefSystems: CoordRefSystems, CRS
+using Meshes: Meshes, Geometry
 
 import FastGaussQuadrature
 import HCubature
 import LinearAlgebra
-import Meshes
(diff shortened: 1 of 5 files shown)


mikeingold:
This partially conflicts with #114 (as-is at the moment). I wanted to shrink the line length of `_units` to fit it into a single-line definition and had this idea, but implementing this for the whole package felt out-of-scope for that PR. If this PR is merged, I'll update the #114 branch to merge it with these changes.

mikeingold:
I'm continuing to consider condensing all of our function definitions to use single-line headers, but this currently would still require splitting somewhere into multiple lines to respect the 92-column soft limit. I'm hoping that this change would also make that change more practical.

JoshuaLampert (approved):

PR #119, opened 2024-10-24
Implement benchmarking with AirspeedVelocity.jl

### LINKED ISSUE #21: Implement benchmark test suite
It would be nice to have a collection of benchmarks for integration methods that could be run to assess potential changes for performance impacts, identify performance regressions and opportunities for improvement, etc.

mikeingold:
## Changes
- Adds a CI Action (duplicate of [this one](https://github.com/MilesCranmer/AirspeedVelocity.jl/blob/master/.github/workflows/benchmark_pr.yml)) that uses AirspeedVelocity.jl to run benchmarks for every new PR
- Adds a benchmark script with a suite containing 14 `@benchmark` entries, testing integrals of `Segment` and `Sphere` (surface), `jacobian`, and `differential`.

Closes #21

### DIFF
--- a/.github/workflows/AirspeedVelocity.yml
+++ b/.github/workflows/AirspeedVelocity.yml
@@ -22,7 +22,7 @@ jobs:
               id: extract-package-name
               run: |
                 PACKAGE_NAME=$(grep "^name" Project.toml | sed 's/^name = "\(.*\)"$/\1/')
-                echo "::set-output name=package_name::$PACKAGE_NAME"
+                echo "package_name=$PACKAGE_NAME" >> $GITHUB_OUTPUT
             - name: Build AirspeedVelocity
               env:
                 JULIA_NUM_THREADS: 2
--- a/.github/workflows/AirspeedVelocity.yml
+++ b/.github/workflows/AirspeedVelocity.yml
@@ -1,7 +1,7 @@
 name: Benchmark a Pull Request
 
 on:
-  pull_request_target:
+  push:
     branches:
       - main
 
--- a/.github/workflows/AirspeedVelocity.yml
+++ b/.github/workflows/AirspeedVelocity.yml
@@ -1,0 +1,78 @@
+name: Benchmark a Pull Request
+
+on:
+  pull_request_target:
+    branches:
+      - master
+
+permissions:
+  pull-requests: write
+
+jobs:
+    generate_plots:
+        runs-on: ubuntu-latest
+
+        steps:
+            - uses: actions/checkout@v2
+            - uses: julia-actions/setup-julia@v1
+              with:
+                version: "1.8"
+            - uses: julia-actions/cache@v1
+            - name: Extract Package Name from Project.toml
+              id: extract-package-name
+              run: |
+                PACKAGE_NAME=$(grep "^name" Project.toml | sed 's/^name = "\(.*\)"$/\1/')
+                echo "::set-output name=package_name::$PACKAGE_NAME"
+            - name: Build AirspeedVelocity
+              env:
+                JULIA_NUM_THREADS: 2
+              run: |
+                # Lightweight build step, as sometimes the runner runs out of memory:
+                julia -e 'ENV["JULIA_PKG_PRECOMPILE_AUTO"]=0; import Pkg; Pkg.add(;url="https://github.com/MilesCranmer/AirspeedVelocity.jl.git")'
+                julia -e 'ENV["JULIA_PKG_PRECOMPILE_AUTO"]=0; import Pkg; Pkg.build("AirspeedVelocity")'
+            - name: Add ~/.julia/bin to PATH
+              run: |
+                echo "$HOME/.julia/bin" >> $GITHUB_PATH
+            - name: Run benchmarks
+              run: |
+                echo $PATH
+                ls -l ~/.julia/bin
+                mkdir results
+                benchpkg ${{ steps.extract-package-name.outputs.package_name }} --rev="${{github.event.repository.default_branch}},${{github.event.pull_request.head.sha}}" --url=${{ github.event.repository.clone_url }} --bench-on="${{github.event.repository.default_branch}}" --output-dir=results/ --tune
+            - name: Create plots from benchmarks
+              run: |
+                mkdir -p plots
+                benchpkgplot ${{ steps.extract-package-name.outputs.package_name }} --rev="${{github.event.repository.default_branch}},${{github.event.pull_request.head.sha}}" --npart=10 --format=png --input-dir=results/ --output-dir=plots/
+            - name: Upload plot as artifact
+              uses: actions/upload-artifact@v2
+              with:
+                name: plots
+                path: plots
+            - name: Create markdown table from benchmarks
+              run: |
+                benchpkgtable ${{ steps.extract-package-name.outputs.package_name }} --rev="${{github.event.repository.default_branch}},${{github.event.pull_request.head.sha}}" --input-dir=results/ --ratio > table.md
+                echo '### Benchmark Results' > body.md
+                echo '' >> body.md
+                echo '' >> body.md
+                cat table.md >> body.md
+                echo '' >> body.md
+                echo '' >> body.md
+                echo '### Benchmark Plots' >> body.md
+                echo 'A plot of the benchmark results have been uploaded as an artifact to the workflow run for this PR.' >> body.md
+                echo 'Go to "Actions"->"Benchmark a pull request"->[the most recent run]->"Artifacts" (at the bottom).' >> body.md
+
+            - name: Find Comment
+              uses: peter-evans/find-comment@v2
+              id: fcbenchmark
+              with:
+                issue-number: ${{ github.event.pull_request.number }}
+                comment-author: 'github-actions[bot]'
+                body-includes: Benchmark Results
+
+            - name: Comment on PR
+              uses: peter-evans/create-or-update-comment@v3
+              with:
+                comment-id: ${{ steps.fcbenchmark.outputs.comment-id }}
+                issue-number: ${{ github.event.pull_request.number }}
+                body-path: body.md
+                edit-mode: replace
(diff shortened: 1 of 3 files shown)


mikeingold:
@JoshuaLampert, this seems to work well on my system and should provide us with benchmarks to quantify performance impacts of each PR. The Action script is taken straight from AirspeedVelocity.jl; I don't see any blatant red flags, but you're more knowledgeable in that domain than I am.

.github/workflows/AirspeedVelocity.yml
```diff
@@ -0,0 +1,78 @@
+name: Benchmark a Pull Request
+
+on:
+  pull_request_target:
+    branches:
+      - master
```
JoshuaLampert, line 6:
```suggestion
      - main
```
We use `main` instead of `master`. I think/hope this is also the reason why this Action didn't run for this PR.

.github/workflows/AirspeedVelocity.yml
```diff
@@ -0,0 +1,78 @@
...
+
+jobs:
+    generate_plots:
+        runs-on: ubuntu-latest
+
+        steps:
+            - uses: actions/checkout@v2
+            - uses: julia-actions/setup-julia@v1
+              with:
+                version: "1.8"
```
JoshuaLampert, line 19:
```suggestion
                version: "1.10"
```
We require at least julia v1.9. So I would go with either v1.9 or v1.10 as we use in the CI tests.

benchmark/benchmarks.jl
```diff
@@ -0,0 +1,49 @@
+using BenchmarkTools
+using LinearAlgebra
+using Meshes
+using MeshIntegrals
+using Unitful
```
JoshuaLampert, line 5:
```suggestion
```
Is Unitful actually needed anywhere explicitly?

JoshuaLampert (changes requested):
This looks quite neat! I would like to see the action already running for this PR. Normally GitHub already runs newly added actions already in the PR, where they are added. So let's see if fixing the `pull_request_target` branch runs the action.

.github/workflows/AirspeedVelocity.yml
```diff
@@ -0,0 +1,78 @@
+name: Benchmark a Pull Request
+
+on:
+  pull_request_target:
```
JoshuaLampert, line 4:
```suggestion
  push:
```
I think `pull_request_target` is not triggered for a `push` event, cf. https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target. I think running the bechmarks for every push is reasonable. We do the same for the CI runs. If I understand correctly, the comment from github-actions will be updated automatically after each new push. Let's see if this triggers the Action.

.github/workflows/AirspeedVelocity.yml
```diff
@@ -0,0 +1,78 @@
+name: Benchmark a Pull Request
+
+on:
+  push:
```
JoshuaLampert, line 4:
```suggestion
  pull_request:
```
Next try...

JoshuaLampert:
Ahh, there we go. Now it runs, but fails. Let me fix that quickly.

.github/workflows/AirspeedVelocity.yml
```diff
@@ -0,0 +1,78 @@
...
+
+permissions:
+  pull-requests: write
+
+jobs:
+    generate_plots:
+        runs-on: ubuntu-latest
+
+        steps:
+            - uses: actions/checkout@v2
```
JoshuaLampert, line 16:
```suggestion
            - uses: actions/checkout@v4
```

.github/workflows/AirspeedVelocity.yml
```diff
@@ -0,0 +1,78 @@
...
+            - uses: julia-actions/setup-julia@v1
```
JoshuaLampert, line 17:
```suggestion
            - uses: julia-actions/setup-julia@v2
```

.github/workflows/AirspeedVelocity.yml
```diff
@@ -0,0 +1,78 @@
...
+              with:
+                version: "1.8"
+            - uses: julia-actions/cache@v1
```
JoshuaLampert, line 20:
```suggestion
            - uses: julia-actions/cache@v2
```

.github/workflows/AirspeedVelocity.yml
```diff
@@ -0,0 +1,78 @@
...
+                echo $PATH
+                ls -l ~/.julia/bin
+                mkdir results
+                benchpkg ${{ steps.extract-package-name.outputs.package_name }} --rev="${{github.event.repository.default_branch}},${{github.event.pull_request.head.sha}}" --url=${{ github.event.repository.clone_url }} --bench-on="${{github.event.repository.default_branch}}" --output-dir=results/ --tune
+            - name: Create plots from benchmarks
+              run: |
+                mkdir -p plots
+                benchpkgplot ${{ steps.extract-package-name.outputs.package_name }} --rev="${{github.event.repository.default_branch}},${{github.event.pull_request.head.sha}}" --npart=10 --format=png --input-dir=results/ --output-dir=plots/
+            - name: Upload plot as artifact
+              uses: actions/upload-artifact@v2
```
JoshuaLampert, line 47:
```suggestion
              uses: actions/upload-artifact@v4
```

.github/workflows/AirspeedVelocity.yml
```diff
@@ -0,0 +1,78 @@
...
+                echo '' >> body.md
+                cat table.md >> body.md
+                echo '' >> body.md
+                echo '' >> body.md
+                echo '### Benchmark Plots' >> body.md
+                echo 'A plot of the benchmark results have been uploaded as an artifact to the workflow run for this PR.' >> body.md
+                echo 'Go to "Actions"->"Benchmark a pull request"->[the most recent run]->"Artifacts" (at the bottom).' >> body.md
+
+            - name: Find Comment
+              uses: peter-evans/find-comment@v2
```
JoshuaLampert, line 65:
```suggestion
              uses: peter-evans/find-comment@v3
```

.github/workflows/AirspeedVelocity.yml
```diff
@@ -0,0 +1,78 @@
...
+              id: fcbenchmark
+              with:
+                issue-number: ${{ github.event.pull_request.number }}
+                comment-author: 'github-actions[bot]'
+                body-includes: Benchmark Results
+
+            - name: Comment on PR
+              uses: peter-evans/create-or-update-comment@v3
```
JoshuaLampert, line 73:
```suggestion
              uses: peter-evans/create-or-update-comment@v4
```

JoshuaLampert (changes requested):
There were some old versions.

.github/workflows/AirspeedVelocity.yml
JoshuaLampert: Ok, now the Action fails because of this.

.github/workflows/AirspeedVelocity.yml
```diff
@@ -0,0 +1,78 @@
...
+                julia -e 'ENV["JULIA_PKG_PRECOMPILE_AUTO"]=0; import Pkg; Pkg.build("AirspeedVelocity")'
+            - name: Add ~/.julia/bin to PATH
+              run: |
+                echo "$HOME/.julia/bin" >> $GITHUB_PATH
+            - name: Run benchmarks
+              run: |
+                echo $PATH
+                ls -l ~/.julia/bin
+                mkdir results
+                benchpkg ${{ steps.extract-package-name.outputs.package_name }} --rev="${{github.event.repository.default_branch}},${{github.event.pull_request.head.sha}}" --url=${{ github.event.repository.clone_url }} --bench-on="${{github.event.repository.default_branch}}" --output-dir=results/ --tune
```
JoshuaLampert, line 41:
```suggestion
                benchpkg ${{ steps.extract-package-name.outputs.package_name }} --rev="${{github.event.repository.default_branch}},${{github.event.pull_request.head.sha}}" --url=${{ github.event.repository.clone_url }} --bench-on="${{github.event.pull_request.head.sha}}" --output-dir=results/ --tune
```
So now it fails because it tries to get the benchmark script from the base branch (i.e. main) instead of from the head branch (i.e. benchmarks), which of course fails because there is no benchmark script on the base branch yet.
IMO, it would be nice to always use the benchmark script from the head branch. If one changes this in a PR, I would assume that this is also used for the benchmark test of this PR and not the old one at main. Let me try if this works.

JoshuaLampert: Indeed, this seems to work :+1:

.github/workflows/AirspeedVelocity.yml
```diff
@@ -0,0 +1,78 @@
...
+            - uses: actions/checkout@v4
+            - uses: julia-actions/setup-julia@v2
+              with:
+                version: "1.10"
+            - uses: julia-actions/cache@v2
+            - name: Extract Package Name from Project.toml
+              id: extract-package-name
+              run: |
+                PACKAGE_NAME=$(grep "^name" Project.toml | sed 's/^name = "\(.*\)"$/\1/')
+                echo "::set-output name=package_name::$PACKAGE_NAME"
```
JoshuaLampert, line 25:
```suggestion
                echo "package_name=$PACKAGE_NAME" >> $GITHUB_OUTPUT
```
This was giving a warning, see, e.g., https://github.com/JuliaGeometry/MeshIntegrals.jl/actions/runs/11516676483/job/32059817164#step:5:6. This updates the command according to https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/.

JoshuaLampert:
Everything runs fine now. Also uploading the plots works and updating the comment after a new push works nicely.

.github/workflows/AirspeedVelocity.yml
```diff
@@ -0,0 +1,78 @@
...
+on:
+  pull_request:
+    branches:
+      - main
+
+permissions:
+  pull-requests: write
+
+jobs:
+    generate_plots:
```
JoshuaLampert, line 12:
```suggestion
    benchmark:
```
IMO the name is not quite appropriate because it does more than generating plots.

mikeingold:
Thanks for getting this working!

JoshuaLampert (approved):

Issue #35, opened 2024-03-27
Automatic differentiation to calculate Jacobians


mikeingold:
The codebase currently uses an inelegant but reasonable finite difference approximation to calculate the Jacobian of a geometry's parametric function.

https://github.com/mikeingold/MeshIntegrals.jl/blob/66ed919150b3774bf09dabab60476869c5514493/src/utils.jl#L5-L15

At time of writing, there didn't appear to be an automatic differentiation package that could compute the same thing out-of-the-box. These parametric functions return a `Point` rather than a simple number. I believe this could be addressed by defining chain rules in Meshes.jl, but I'm open to other solutions.

mikeingold:
Meshes.jl now uses Zygote.jl as an AD-of-choice. Ideally, all `Geometry`s in Meshes.jl would have diff rules established so that this Jacobian function would flow from that.

mikeingold:
Merged with #37

mikeingold:
Since `differential` was generalized in #101, I'm going to un-merge this Issue from #37.

JoshuaLampert:
In #115 we agreed that it would be best to implement the AD support over in Meshes.jl (we could also think about transferring the issue?).
I haven't tested anything AD related in the context of Meshes.jl and MeshIntegrals.jl (I only have a bit of experience using Zygote.jl, ForwardDiff.jl and Enzyme.jl in other contexts), but I see that the fact that the parametric functions of the geometries return `Point`s complicates the issue and probably necessitates implementing custom AD rules via ChainRules.jl.
> I’m also not sure if writing these directly for parametric functions is the best path, or if we can write them for lower-level functions and have that support roll up automatically.

I would opt for minimizing the number of rules to implement, i.e. not needing to implement a rule for each geometry, but rather write something more low level. This seems to be easier to maintain and extend.
These are just my initial thought without doing any experimenting myself. My familiarity with ChainRules.jl is pretty low. So I'm not sure if this approach is feasible.
Getting some advice from @kylebeggs would be awesome.

mikeingold:
> ... but I see that the fact that the parametric functions of the geometries return `Point`s complicates the issue and probably necessitates implementing custom AD rules via ChainRules.jl.

[...]

PR #122, opened 2024-10-26
WIP: Use a new DifferentiationMethod to dispatch for derivative types

mikeingold:
## Proposed Changes
- Adds a `DifferentiationMethod` abstract type to communicate derivative calculation method
  - `FiniteDifference(ε=1e-6)` is the only method yet defined
  - Added commented-out placeholders for `EnzymeAD` and `ZygoteAD`
- Changes function signatures for `differential` and `jacobian` to 3-argument: `(geometry, coords, ::DifferentiationMethod)`

TODO
- Change `integral` calls

Ref #35

### DIFF
--- a/src/differentiation.jl
+++ b/src/differentiation.jl
@@ -80,17 +80,18 @@ end
 ################################################################################
 
 """
-    differential(geometry, ts)
+    differential(geometry, ts, method=FiniteDifference())
 
 Calculate the differential element (length, area, volume, etc) of the parametric
 function for `geometry` at arguments `ts`.
 """
 function differential(
         geometry::Geometry,
-        ts::V
+        ts::V,
+        method::DifferentiationMethod = FiniteDifference()
 ) where {V <: Union{AbstractVector, Tuple}}
     # Calculate the Jacobian, convert Vec -> KVector
-    J = jacobian(geometry, ts)
+    J = jacobian(geometry, ts, method)
     J_kvecs = Iterators.map(_kvector, J)
 
     # Extract units from Geometry type
--- a/src/differentiation.jl
+++ b/src/differentiation.jl
@@ -1,5 +1,21 @@
 ################################################################################
-#                     Jacobian and Differential Elements
+#                               JacobianMethods
+################################################################################
+
+abstract type JacobianMethod end
+
+struct FiniteDifference{T<:AbstractFloat} <: JacobianMethod
+    ε::T
+end
+
+# struct EnzymeAD <: JacobianMethod
+# end
+
+# struct ZygoteAD <: JacobianMethod
+# end
+
+################################################################################
+#                                  Jacobian
 ################################################################################
 
 """
--- a/src/differentiation.jl
+++ b/src/differentiation.jl
@@ -2,16 +2,22 @@
 #                               JacobianMethods
 ################################################################################
 
-abstract type JacobianMethod end
+abstract type DifferentiationMethod end
 
-struct FiniteDifference{T<:AbstractFloat} <: JacobianMethod
+"""
+    FiniteDifference(ε)
+
+Use a finite-difference approximation method to calculate derivatives with a
+step size of `ε`.
+"""
+struct FiniteDifference{T<:AbstractFloat} <: DifferentiationMethod
     ε::T
 end
 
-# struct EnzymeAD <: JacobianMethod
+# struct EnzymeAD <: DifferentiationMethod
 # end
 
-# struct ZygoteAD <: JacobianMethod
+# struct ZygoteAD <: DifferentiationMethod
 # end
 
 ################################################################################
--- a/src/differentiation.jl
+++ b/src/differentiation.jl
@@ -14,6 +14,9 @@ struct FiniteDifference{T<:AbstractFloat} <: DifferentiationMethod
     ε::T
 end
 
+# If ε not specified, default to 1e-6
+FiniteDifference() = new(1e-6)
+
 # struct EnzymeAD <: DifferentiationMethod
 # end
 
--- a/src/differentiation.jl
+++ b/src/differentiation.jl
@@ -40,8 +40,7 @@ function jacobian end
 function jacobian(
         geometry::Geometry,
         ts::V,
-        fd::FiniteDifference;
-        ε = 1e-6
+        fd::FiniteDifference
 ) where {V <: Union{AbstractVector, Tuple}}
     Dim = Meshes.paramdim(geometry)
     if Dim != length(ts)
--- a/src/differentiation.jl
+++ b/src/differentiation.jl
@@ -72,36 +72,6 @@ function jacobian(
     return ntuple(n -> ∂ₙr(ts, n), Dim)
 end
 
-function jacobian(
-        bz::Meshes.BezierCurve,
-        ts::V
-) where {V <: Union{AbstractVector, Tuple}}
-    t = only(ts)
-    # Parameter t restricted to domain [0,1] by definition
-    if t < 0 || t > 1
-        throw(DomainError(t, "b(t) is not defined for t outside [0, 1]."))
-    end
-
-    # Aliases
-    P = bz.controls
-    N = Meshes.degree(bz)
-
-    # Ensure that this implementation is tractible: limited by ability to calculate
-    #   binomial(N, N/2) without overflow. It's possible to extend this range by
-    #   converting N to a BigInt, but this results in always returning BigFloat's.
-    N <= 1028 || error("This algorithm overflows for curves with ⪆1000 control points.")
-
-    # Generator for Bernstein polynomial functions
-    B(i, n) = t -> binomial(Int128(n), i) * t^i * (1 - t)^(n - i)
-
-    # Derivative = N Σ_{i=0}^{N-1} sigma(i)
-    #   P indices adjusted for Julia 1-based array indexing
-    sigma(i) = B(i, N - 1)(t) .* (P[(i + 1) + 1] - P[(i) + 1])
-    derivative = N .* sum(sigma, 0:(N - 1))
-
-    return (derivative,)
-end
-
 ################################################################################
 #                          Differential Elements
 ################################################################################
--- a/src/differentiation.jl
+++ b/src/differentiation.jl
@@ -25,19 +25,22 @@ end
 ################################################################################
 
 """
-    jacobian(geometry, ts; ε=1e-6)
+    jacobian(geometry, ts, method)
 
-Calculate the Jacobian of a geometry at some parametric point `ts` using a simple
-finite-difference approximation with step size `ε`.
+Calculate the Jacobian of a geometry's parametric function at some point `ts`
+using a particular differentiation method.
 
 # Arguments
 - `geometry`: some `Meshes.Geometry` of N parametric dimensions
 - `ts`: a parametric point specified as a vector or tuple of length N
-- `ε`: step size to use for the finite-difference approximation
+- `method`: the desired `DifferentiationMethod`
 """
+function jacobian end
+
 function jacobian(
         geometry::Geometry,
-        ts::V;
+        ts::V,
+        fd::FiniteDifference;
         ε = 1e-6
 ) where {V <: Union{AbstractVector, Tuple}}
     Dim = Meshes.paramdim(geometry)
@@ -46,7 +49,7 @@ function jacobian(
     end
 
     T = eltype(ts)
-    ε = T(ε)
+    ε = T(fd.ε)
 
     # Get the partial derivative along the n'th axis via finite difference
     #   approximation, where ts is the current parametric position


src/differentiation.jl
```diff
@@ -1,30 +1,64 @@
...
+# struct ZygoteAD <: DifferentiationMethod
+# end
+
+################################################################################
+#                                  Jacobian
 ################################################################################
 
 """
-    jacobian(geometry, ts; ε=1e-6)
+    jacobian(geometry, ts, method)
```
JoshuaLampert, line 31:
```suggestion
    jacobian(geometry, ts, method=FiniteDifference())
```

src/integral.jl
```diff
@@ -3,7 +3,7 @@
 ################################################################################
 
 """
-    integral(f, geometry[, rule]; FP=Float64)
+    integral(f, geometry[, rule]; dt=FiniteDifference(), FP=Float64)
```
JoshuaLampert, line 6: What does `dt` stand for? Can this maybe be named a bit more self-explanatory? When I read `dt` I immediately think of a time step size for a numerical solver of an ODE (could also just be my background), but I think generally something like `diff_method` or something along those lines would be easier to understand what this does.

JoshuaLampert (commented):
I would also `export FiniteDifference`. We could also think about a backend `Analytical` or similar, which computes the analytical derivative, which would currently be only supported for a `BezierCurve`, but that's fine IMO. What do you think?

src/integral.jl
mikeingold: I mostly just wanted to get something implemented for testing and figured we could debate the name. I was thinking in terms of like $\frac{\text{d}}{\text{d}t}~\text{geometry}(t)$, but that’s a fair point.

JoshuaLampert:
This is breaking because, e.g., `jacobian(geometry, ts; ε=1e-7)`, doesn't work anymore.

mikeingold:
> This is breaking because, e.g., `jacobian(geometry, ts; ε=1e-7)`, doesn't work anymore.

Yes. I’ve been planning for this to be part of a `v0.16` release. That change is breaking and this new support structure around specifying diff type is a non-trivial feature update anyway.

JoshuaLampert:
Yeah, I'm fine with that :+1:

src/integral.jl
mikeingold: I just changed `dt` to `diff_method`

JoshuaLampert: Thanks!

mikeingold:
> We could also think about a backend `Analytical` or similar, which computes the analytical derivative, which would currently be only supported for a `BezierCurve`, but that's fine IMO. What do you think?

My feeling is that if an analytic solution exists then using that is generally the superior option. In a case like `BezierCurve` there’s still a distinct `jacobian` definition, but for a number of other types like `Line` and `Triangle` there’s only an implicit analytic definition buried in the solution itself, so `jacobian` is not even called. 

Based on all of that, I think the most consistent solution on balance is probably just to default to analytic solutions for the minority of cases where they’re going to exist, even when a specific method is requested.

JoshuaLampert:
So the timings in the benchmark Action (despite the fact that it's failing) show that the analytical solution for the Jacobian of the Bezier curve is really crucial for performance. Thus, we definitely need to keep it.
I see two possible solutions:
1. Similar to how it is now, i.e. for those cases where we have an analytical solution implemented, we always use it even if there is another method like `FiniteDifference` requested. This means we would add back the analytical solution for the `BezierCurve` like the currently commented code you prepared and remove the `diff_method` argument from the call of `jacobian`. IIUC, this is your preferred version.
2. As I suggested above we add another backend `Analytical` and add all analytical derivatives we have (`BezierCurve`, `Line`, `Plane` etc.) as `jacobian(geometry, ts, diff_method::Analytical)` to differentiation.jl. Then we would set the default option for `diff_method` to `Analytical()` for all `geometry`s, where we have them and to `FiniteDifference() ` otherwise (we would need a smart way to do that; maybe something like `has_analytical_derivative(::Geometry) = false`, `has_analytical_derivative(::BezierCurve) = true)` etc. and then set the default `diff_method = has_analytical_derivative(geometry) ? Analytical() : FiniteDifference()`) 

I think I prefer the second option for the following reasons:
a) You _can_ use other backends also in the case of the geometries where you have analytical solutions if you want. Of course this is generally not a good idea for performance (and accuracy in the case of FD), but you _can_ do it if you want to investigate the influence of the underlying backend for instance. It could also help us to test AD with these `geometry`s in the future. Still, you have the best default behaviour.
b) It could be confusing if you request a backend like `FiniteDifference()` or some of the AD backends in the future, but this is not even used.
c) What if we decide to implement another analytical derivative in the future, say for a `geometry`, which does not have a specialization? If we follow approach 1. we would need to add a specialization of `integral` just to call `jacobian(geometry, ts)` (to call the new analytical derivative) instead of the generic `jacobian(geometry, ts, diff_method)`. On the other hand, if we use approach 2. we would only need to add the new `jacobian(geometry, ts, diff_method::Analytical)` and dispatch will do the rest to call the correct `jacobian` method.

I also think option 2. is consistent in the sense that we always call `jacobian(geometry, ts, diff_method)` (and not some `jacobian(geometry, ts)`, which is defined somewhere else as the other `jacobian` methods), have an analytical solution as default if available, and have FD as default if there is no analytical solution. What do you think, @mikeingold?

JoshuaLampert:
The `Key "25" not found` error will probably be gone once we use the analytical derivative for `BezierCurve`  again because as I understand from the linked issue in AirspeedVelocity.jl this error comes from the fact that the Bezier curve tests are only evaluated once (because they take so long) leading to the non-existence of standard deviation in these cases, which in turn let's Airspeedvelocity.jl fail.

mikeingold:
> The `Key "25" not found` error will probably be gone once we use the analytical derivative for `BezierCurve` again because as I understand from the linked issue in AirspeedVelocity.jl this error comes from the fact that the Bezier curve tests are only evaluated once (because they take so long) leading to the non-existence of standard deviations in these cases, which in turn let's AirspeedVelocity.jl fail.

Agreed. I checked the BenchmarkTools.jl docs and it looks like I misunderstood the role of the evals keyword. We should probably just remove all of them and leave the default value of 1.

mikeingold:
@JoshuaLampert I've been thinking hard about the API here and the good points you've made. I think I've got a new plan that's somewhat of a middle ground between the two options you presented. I'm on mobile right now but will try to explain my thoughts better tomorrow.

mikeingold:
I wrote out an applicability table that outlines the support we'd be able to provide for particular differentiation methods. For clarity, I dropped from the support matrix the geometries that are currently-unsupported (e.g. `Pyramid`) or simply composite types (e.g. `Rope`).

**Hard-Coded Analytic:** These geometries are outliers with `specializations` that perform custom non-linear domain transformations and so don't rely on the standard `differential` or `jacobian` at all.

**Analytic:** `BezierCurve` is the only geometry currently supported. For the few cases where a Hard-Coded Analytic solution is used, specifying this type would be redundant/confirmation. It's possible for us to define analytic methods for almost all others types. However, these may be fragile to changes in the upstream Meshes parametric functions, and performance will likely be comparable to AD methods in most cases. This leads to a **hard question**: should we even be defining these?

**FiniteDifference:** This isn't the most elegant or performant solution, but it's the current default (for everything but the Hard-Coded Analytic cases and `BezierCurve`) because it mostly Just Works.

**Auto\*:** All but the Hard-Coded Analytic cases should be supported by `AutoEnzyme` and (hopefully) `AutoZygote` once they are implemented.

| `Meshes.Geometry` | Hard-Coded Analytic | `Analytic` | `FiniteDifference` | `Auto*` |
|----------|----------------|---------------|---------------------|-----|
| `Ball`              | 🛑 | 🎗️ | ✅ | 🎗️ |
| `BezierCurve`       | 🛑 | ✅ | ✅ | 🎗️ |
| `Box`               | 🛑 | 🎗️ | ✅ | 🎗️ |
| `Circle`            | 🛑 | 🎗️ | ✅ | 🎗️ |
| `Cone`              | 🛑 | 🎗️ | ✅ | 🎗️ |
| `Cylinder`          | 🛑 | 🎗️ | ✅ | 🎗️ |
| `Disk`              | 🛑 | 🎗️ | ✅ | 🎗️ |
| `Ellipsoid`         | 🛑 | 🎗️ | ✅ | 🎗️ |
| `Line`              | ✅ | ⬅️ | 🛑 | 🛑 |
| `ParaboloidSurface` | 🛑 | 🎗️ | ✅ | 🎗️ |
| `ParameterizedCurve` | 🛑 | 🛑 | ✅ | 🎗️ |
| `Plane`             | ✅ | ⬅️ | 🛑 | 🛑 |
| `Quadrangle`        | 🛑 | 🎗️ | ✅ | 🎗️ |
| `Ray`               | ✅ | ⬅️ | 🛑 | 🛑 |
| `Segment`           | 🛑 | 🎗️ | ✅ | 🎗️ |
| `Sphere`            | 🛑 | 🎗️ | ✅ | ✅ |
| `Tetrahedron`       | ✅ | ⬅️ | 🛑 | 🛑 |
| `Triangle`          | ✅ | ⬅️ | 🛑 | 🛑 |
| `Torus`             | 🛑 | 🎗️ | ✅ | 🎗️ |

| Symbol | Applicability Level |
|--------|---------|
| ✅ | Currently supported |
| ⬅️ | `Analytic` consistent with hard-coded methods.
| 🎗️ | Possible to implement |
| 🛑 | Not supported |

In terms of dealing with the Hard-Coded cases, one possible solution might be to simply couch the language in the documentation to indicate something like ... "`diff_method` specifies the technique to use when an analytic solution is not available".

If we decide to implement `Analytic()` then, as of this PR, the kwarg default order of precedence should probably be `Analytic()` > `FiniteDifference()`. In the future, if/when we get `Auto*` support, then we can evaluate where how it performs against `Analytic` and maybe relegate `FiniteDifference` to a by-request-only option.

mikeingold:
`:%s/Analytic/Analytical`

mikeingold:
I need to update the summary at the top based on the latest direction, but I think we're nearly there on this PR.

mikeingold:
Weird. Benchmark results don't seem to have been fixed for `BezierCurve` integrals.

mikeingold:
Fixed. Looks like it was a dispatch issue.

mikeingold:
Update: I think I'm pretty much done with the code here. I just need to verify code coverage and finish updating the documentation now.

src/differentiation.jl
```diff
@@ -1,30 +1,87 @@
...
+
+################################################################################
+#                                  Jacobian
+################################################################################
+
+"""
+    jacobian(geometry, ts[, diff_method])
+
+Calculate the Jacobian of a `geometry`'s parametric function at some point `ts`.
+Optionally, direct the use of a particular `differentiation method`; by default
```
JoshuaLampert, line 55:
```suggestion
Optionally, direct the use of a particular differentiation method `diff_method`; by default
```

src/differentiation.jl
```diff
@@ -1,30 +1,87 @@
...
 
 # Arguments
 - `geometry`: some `Meshes.Geometry` of N parametric dimensions
 - `ts`: a parametric point specified as a vector or tuple of length N
-- `ε`: step size to use for the finite-difference approximation
+- `diff_method`: the desired `DifferentiationMethod` to use
 """
+function jacobian end
+
+function jacobian(
```
JoshuaLampert, line 66:
```suggestion
"""
function jacobian(
```

src/differentiation.jl
```diff
@@ -1,30 +1,87 @@
...
+) where {G <: Geometry, V <: Union{AbstractVector, Tuple}}
+    return jacobian(geometry, ts, _default_method(G))
+end
+
 function jacobian(
         geometry::Geometry,
-        ts::V;
-        ε = 1e-6
+        ts::V,
+        diff_method::FiniteDifference
```
JoshuaLampert, line 76: Why not specifying the default here `diff_method::FiniteDifference = _default_method(Geometry)` and delete the method above?

src/differentiation.jl
```diff
@@ -48,52 +105,30 @@ function jacobian(
...
 ################################################################################
 
 """
-    differential(geometry, ts)
+    differential(geometry, ts[, diff_method])
 
 Calculate the differential element (length, area, volume, etc) of the parametric
-function for `geometry` at arguments `ts`.
+function for `geometry` at arguments `ts`. Optionally, direct the use of a
+particular `differentiation method`; by default use analytic solutions where
```
JoshuaLampert, line 117:
```suggestion
particular differentiation method `diff_method`; by default use analytic solutions where
```

src/differentiation.jl
```diff
@@ -48,52 +105,30 @@ function jacobian(
...
+possible and finite difference approximations otherwise.
+
+# Arguments
+- `geometry`: some `Meshes.Geometry` of N parametric dimensions
+- `ts`: a parametric point specified as a vector or tuple of length N
+- `diff_method`: the desired `DifferentiationMethod` to use
```
JoshuaLampert, line 123:
```suggestion
- `diff_method`: the desired [`DifferentiationMethod`](@ref) to use
```

src/differentiation.jl
```diff
@@ -1,30 +1,87 @@
...
+Calculate the Jacobian of a `geometry`'s parametric function at some point `ts`.
+Optionally, direct the use of a particular `differentiation method`; by default
+use analytic solutions where possible and finite difference approximations
+otherwise.
 
 # Arguments
 - `geometry`: some `Meshes.Geometry` of N parametric dimensions
 - `ts`: a parametric point specified as a vector or tuple of length N
-- `ε`: step size to use for the finite-difference approximation
+- `diff_method`: the desired `DifferentiationMethod` to use
```
JoshuaLampert, line 62:
```suggestion
- `diff_method`: the desired [`DifferentiationMethod`](@ref) to use
```

src/differentiation.jl
```diff
@@ -1,30 +1,87 @@
...
 
 """
-    jacobian(geometry, ts; ε=1e-6)
+    DifferentiationMethod
 
-Calculate the Jacobian of a geometry at some parametric point `ts` using a simple
-finite-difference approximation with step size `ε`.
+A category of types used to specify the desired method for calculating derivatives.
+Derivatives are used to form Jacobian matrices when calculating the differential
+element size throughout the integration region.
```
JoshuaLampert, line 10:
```suggestion
element size throughout the integration region.

See also [`FiniteDifference`](@ref), [`Analytical`](@ref).
```

src/integral.jl
```diff
@@ -3,7 +3,7 @@
 ################################################################################
 
 """
-    integral(f, geometry[, rule]; FP=Float64)
+    integral(f, geometry[, rule]; diff_method, FP)
```
JoshuaLampert, line 6:
```suggestion
    integral(f, geometry[, rule]; diff_method=_default_method(geometry), FP=Float64)
```
Why not specify the default arguments already here?

JoshuaLampert (changes requested):
This is on a good way. Some intermediate feedback mostly concerning documentation.

kylebeggs:
Is the API pretty much set here so I can adopt it for the Enzyme work in #35 ?

mikeingold:
> Is the API pretty much set here so I can adopt it for the Enzyme work in #35 ?

Yes. At this point in #122 I don’t foresee any more structural changes coming.

JoshuaLampert:
If you @kylebeggs have any suggestions or thoughts about the interface as prepared in this PR, feel free to share them. I (and I guess @mikeingold too) would be interested in a third opinion :smiley:

src/differentiation.jl
mikeingold: My goal here was to avoid conditional blocks by leveraging dispatch to define three-arg methods specific to each DM type. If the default DM were always `FiniteDifference` then I think this note would be a good idea, but the default could be any DM type, which could require dispatch to a different method than this one. In that sense, I think it makes the code easier to reason about (and more flexible moving forward) when the sole two-argument method is simply a wrapper over the three-argument with an inferred `diff_method` argument.

JoshuaLampert: Makes sense.

mikeingold:
@kylebeggs are you planning to implement with an extension or directly in the package? Early on my thinking went directly to extension, but I’m not sure if defining sub-types from an extension module outside the main namespace is problematic or not.

mikeingold:
I can foresee some confusion about why an integration package needs differentiation methods, so I’m using this as an excuse to finally close #62. I’m adding an explainer page to the documentation with a primer on how we use differential forms to solve integrals, and how that requires Jacobians to be calculated.

src/utils.jl
```diff
@@ -14,6 +14,31 @@ function _error_unsupported_combination(geometry, rule)
...
+end
+
+_guarantee_analytical(G, ::Analytical) = nothing
+
+# Return whether a geometry type has jacobian methods defined
+_has_analytical(::Type{G}) where {G <: Geometry} = false
+_has_analytical(g::G) where {G <: Geometry} = _has_analytical(G)
+
+# Return the default DifferentiationMethod instance for a particular geometry type
+function _default_method(
```
kylebeggs, line 33: I think we should add the floating point precision as an arg for this method because that would inform a better choice for the step size for `FiniteDifference` rather than always 1e-6.

kylebeggs:
@mikeingold I initially was thinking extension as well. I don't have too much experience with extensions, so I'm not sure if that'll present an issue, but my guess is not.

src/utils.jl
mikeingold: Ah. I hadn’t really put much consideration into how `FP` choice flowed into differential methods. Right now in `jacobian` we have a heuristic that the FP type is the same as the coordinate type, but I’ll have to think about this some more.

JoshuaLampert:
@kylebeggs is your idea to first add the AD stuff to MeshIntegrals.jl and once that is settled move it over to Meshes.jl?

kylebeggs:
@JoshuaLampert For Enzyme, yes, for the sake of getting this package AD support in some form as soon as possible. I was planning on adding Zygote/ChainRules support to Meshes.jl directly.

JoshuaLampert:
That sounds reasonable. Meshes.jl already depends on Zygote implicitly via CoordinateRefSystems.jl. So it makes sense to add it there. Regarding Enzyme I would prefer a package extension in MeshIntegrals.jl. It should be quite easy to set up. If you need help regarding this, I can try to help.

src/utils.jl
mikeingold: I don't know that there's a perfect solution for this since `FP` can equally be inferred from the `FiniteDifference` instance or from coordinates, but I've updated the `jacobian` API to better capture the `eltype` of `ts` and enforced that as the `FP` type.

kylebeggs: I think that is fine, and now for this method we can do something like
```julia
function _default_method(g::Type{G}, T::DataType) where {G <: Geometry}
    return _has_analytical(G) ? Analytical() : FiniteDifference(T)
end

FiniteDifference(T::DataType) = T isa Float32 ? FiniteDifference(1e-3) : FiniteDifference()
```
which should be a better choice when working with Float32.

mikeingold:
Docs are done, just need to figure out what seems to be an issue with dispatch or type params.

mikeingold:
Damn. Looks like you can't use keyword argument information in other kwarg defaults?

mikeingold:
Agreed. I went back to single-arg and will go forward with inferencing `FP` based on `eltype(ts)`

mikeingold:
Looks like the improvement came with commit #f845289 where the `jacobian`
type params for `ts` changed from `Tuple` to `NTuple`. I'm not totally sure
why that's worth a 3x speed boost, but I'll take it.

mikeingold:
> Looks like the improvement came with commit #f845289 where the `jacobian` type params for `ts` changed from `Tuple` to `NTuple`. I'm not totally sure why that's worth a 3x speed boost, but I'll take it.

I suspect this is simply because `NTuple` additionally guarantees that all elements are of identical type.

mikeingold:
### Status update

Remaining TODO:
- We're currently failing an Aqua test for unbound type parameters but I'm not entirely clear on why.
- Documentation build is apparently failing due to presence of unicode.
- I have one more minor tweak to the `FiniteDifference` version of `jacobian` that I want to try. I suspect we might get some additional performance out of it.

kylebeggs:
> I suspect this is simply because NTuple additionally guarantees that all elements are of identical type.
That is something I haven't thought about, but it makes sense! Thanks for all the work on this, it is shaping up wonderfully.

mikeingold:
The failing Aqua test for unbound params in `jacobian` is a puzzle for me. The code functions correctly, but technically `N` is unbound if `v` is a vector and not an `NTuple`. Not sure how else to express this in a way that still works nicely but also passes the test.

https://julialang.zulipchat.com/#narrow/channel/225542-helpdesk/topic/Aqua.20unbound.20type.20parameter.20on.20Union

docs/src/how_it_works.md
```diff
@@ -0,0 +1,77 @@
...
+\mathbf{J}_f = \begin{bmatrix} \bar{\text{d}t_1} & \bar{\text{d}t_2} & \bar{\text{d}t_3} \end{bmatrix}
+```
+where
+```math
+\bar{\text{d}t_n} = \frac{\partial}{\partial t_n} ~ \text{g}(t_1,~t_2,~t_3)
+```
+
+Each of these partial derivatives is a vector representing the direction that changing each parametric function argument will move the resultant point. The differential element ($E$) size is then calculated using geometric algebra as the magnitude of the exterior product ($\wedge$) of these three vectors.
+```math
+E(t_1,~t_2,~t_3) = \left\| \bar{\text{d}t_1} \wedge \bar{\text{d}t_2} \wedge \bar{\text{d}t_3} \right\|
```
JoshuaLampert, line 69: Maybe a stupid question, which I also could have asked before:
Isn't this not the absolute value of the determinant of the Jacobian of g? For example in 3D:
```julia
julia> J = rand(3, 3)
3×3 Matrix{Float64}:
 0.000905113  0.494132  0.449875
 0.417646     0.122622  0.532629
 0.408989     0.498596  0.854013

julia> abs(det(J))
0.0023697470337338046

julia> abs((J[:, 1] × J[:, 2]) ⋅ J[:, 3])
0.002369747033733771

julia> using CliffordNumbers

julia> J_kvecs = [CliffordNumbers.KVector{1, VGA(3)}(J[:, 1]...), CliffordNumbers.KVector{1, VGA(3)}(J[:, 2]...), CliffordNumbers.KVector{1, VGA(3)}(J[:, 3]...)]
3-element Vector{KVector{1, VGA(3), Float64, 3}}:
 KVector{1, VGA(3), Float64}(0.0009051134514089787, 0.41764622933744966, 0.4089885721993082)
   KVector{1, VGA(3), Float64}(0.49413189336803354, 0.12262234379158543, 0.4985961890818985)
    KVector{1, VGA(3), Float64}(0.44987481501348114, 0.5326286546861126, 0.8540130191622195)

julia> foldl(∧, J_kvecs)
1-element KVector{3, VGA(3), Float64}:
0.002369747033733785e₁e₂e₃
```
which means all of the three methods (`abs(det(J))` as I propose, the explicit dimension-dependent version we used before switching to CliffordNumbers.jl and the CliffordNumbers.jl version we use now) give the same result. Similar for 2D:
```julia
julia> J2 = rand(2, 2)
2×2 Matrix{Float64}:
 0.572102  0.068292
 0.937427  0.505511

julia> abs(det(J2))
0.2251849308605629

julia> using Meshes

julia> norm(Vec(J2[:, 1]...) × Vec(J2[:, 2]...))
0.22518493086056293 m^2

julia> J_kvecs = [CliffordNumbers.KVector{1, VGA(2)}(J2[:, 1]...), CliffordNumbers.KVector{1, VGA(2)}(J2[:, 2]...)]
2-element Vector{KVector{1, VGA(2), Float64, 2}}:
  KVector{1, VGA(2), Float64}(0.5721015978943291, 0.9374270118081014)
 KVector{1, VGA(2), Float64}(0.06829195845197433, 0.5055110114482464)

julia> foldl(∧, J_kvecs)
1-element KVector{2, VGA(2), Float64}:
0.22518493086056293e₁e₂
```

Do we actually need CliffordNumbers.jl? If `abs(det(J))` gives the same result, this would reduce complexity and resembles the integration by substitution formula (https://en.wikipedia.org/wiki/Integration_by_substitution#Substitution_for_multiple_variables) closely, which nothing really else is the whole story about parametric functions about, isn't it? I mean, the CliffordNumbers.jl solution is really elegant, but so is `abs(det(J))`, isn't it? And it involves less complicated maths, such that people with knowledge from basic multidimensional analysis and integrals (change of variables in multi-D) can understand what happens without needing to have a deeper knowledge about Clifford numbers. Or am I missing something?
Sorry for getting a bit off-topic of this PR, but since this PR also contains the documentation part, I wanted to mention it here.

src/differentiation.jl
```diff
@@ -1,42 +1,99 @@
...
+
+See also [`FiniteDifference`](@ref), [`Analytical`](@ref).
+
+See also [`FiniteDifference`](@ref), [`Analytical`](@ref).
```
JoshuaLampert, line 14:
```suggestion
```

src/specializations/Line.jl
```diff
@@ -24,32 +27,38 @@ function integral(
...
+) where {F <: Function, DM <: DifferentiationMethod, T <: AbstractFloat}
+    _guarantee_analytical(Meshes.Line, diff_method)
+
     # Normalize the Line s.t. line(t) is distance t from origin point
     line = Meshes.Line(line.a, line.a + Meshes.unormalize(line.b - line.a))
 
     # Integrate f along the Line
     domainunits = _units(line(0))
-    integrand(t) = f(line(t)) * domainunits
+    integrand(t) = f(line(t)) * domainunits       # TODO differential
```
JoshuaLampert, line 49: There is an unresolved TODO here.

JoshuaLampert (changes requested):
I have a few smaller suggestions and I left a question for discussion.

docs/src/how_it_works.md
```diff
@@ -0,0 +1,77 @@
+# How it Works (By Example)
+
+## Example Problem
+
+Let $f$ be a function of position ($\bar{r}$)
+```julia
```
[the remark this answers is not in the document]
mikeingold: In the short term, this page is definitely imperfect but is mostly meant to just get something at least mildly useful written down. I think my prior statement that it closes #62 was probably optimistic (though I've come to dislike these kinds of Issues with non-specific closing criteria).

My early goal was to use Makie to output `viz` figures of the `ball` and of the Jacobian vectors at some point on the ball. I didn't quite find a version that I felt looked great but didn't want to continue to hold up implementing this PR over it. How about we track this as a new Issue (enhancement) with ideas for future improvements?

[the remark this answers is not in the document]
JoshuaLampert: Yeah, that's fine with me. This PR is already big enough and the documentation is also kind of orthogonal to the original purpose of the PR.

docs/src/how_it_works.md
mikeingold: This PR is mainly focused on implementing `DifferentiationMethod`, and the doc updates were meant to shed some light on why derivatives are even being calculated.

That said... I actually hadn't realized that `abs(det(J::Matrix))` was mathematically equivalent to the current `abs(foldl(∧, J::NTuple{N,KVector})`. I'm really happy with the elegance of #114, but if we can develop a determinant-based solution that is similarly elegant and performant, then it's probably superior in practice for the reasons you've mentioned. I suspect that solution could look something like
```julia
using StaticArrays

function _SMatrix(
    vecs::NTuple{N, T}
) where {N, T <: Meshes.Vec}
    return SMatrix{N, N}((vecs...)...)
end

differential(geometry, ts, diff_method) = jacobian(geometry, ts, diff_method) |> _SMatrix |> det |> abs
```

I propose we document this is a new enhancement Issue and, if a PR for it pans out as expected, we could tweak this docs page again to explain the new method.

mikeingold:
The latest commit fixes the unbound parameters issue, so hopefully the newfound performance improvement is also maintained.

src/specializations/Line.jl
mikeingold: I just made a commit removing this comment; just waiting on the current Actions to complete. I previously made an effort to tweak all the other special-case methods to make it obvious which part of the integrand expression was essentially just a baked-in `differential(geometry, ts, ::Analytical)`. The `GaussKronrod` methods, however, turn out to be special cases because no domain transform was required since `quadgk` can integrate on $(-\infty, \infty)$ directly.

docs/src/how_it_works.md
JoshuaLampert: I just realized that the Jacobian doesn't need to be square, i.e. it is not square if the `paramdim` of the `geometry` is not its `embeddim`. In this case the determinant solution would not work, while the `foldl(∧, J_kvecs)` works also in this case.

mikeingold: Damn, I was pretty excited about this but you're correct. E.g. a `Segment` embedded in 3D space would have a Jacobian tuple consisting of a single 3D `Meshes.Vec`.

mikeingold: Maybe still worth documenting the idea along with the drawback, just in case we want to revisit it later with fresh ideas.

mikeingold:
Looks like performance was maintained. I think we might be essentially done with this one, finally.

docs/src/how_it_works.md
JoshuaLampert: Agreed, but it's also interesting to see that the geometric algebra solution seems to be kind of a generalization of the classical substitution formula for transformations mapping between domains of different dimensions. Definitely, learned something by thinking about this :smiley:

mikeingold: I'm just now reading that link and was surprised to see how directly-comparable that first theorem was to what we've implemented. I've definitely learned a lot along the way while building this package.

mikeingold:
The OP post has been updated with a summary of all changes. Bumping one last time for any remaining review notes/suggestions.

docs/src/how_it_works.md
```diff
@@ -0,0 +1,77 @@
...
+```math
+\iiint f(\bar{r}) ~ \text{d}V
+```
+
+Integrals like this are often solved manually by selecting an appropriate coordinate system and limits that neatly represent the integration domain, e.g.
+```math
+\int_0^{\pi} \int_0^{2\pi} \int_0^{5} f(\bar{r}) ~ \text{d}\rho~\text{d}\theta~\text{d}\phi
+```
+
+This works great for simple geometries, but requires integration code that is geometry-specific. This package, by contrast, leverages parametric functions defined in Meshes.jl and differential forms to define integral methods that are general solutions for all geometries.
```
JoshuaLampert, line 30: Sorry for bringing this up again, but is it really "by contrast"? Aren't we essentially doing exactly this? I mean parametric functions are basically "selecting an appropriate coordinate system", aren't they? It's also "geometry-specific" because we need the geometry-specific parametric function (only that this is defined in Meshes.jl). Having a generic interface for the parametric functions (e.g. the fact that they always (well, with exceptions like `Line` etc.) map a $[0, 1]^d$ hypercube) allows us to define integral methods in MeshIntegrals.jl that are general for all geometries. Your limits from the coordinate system $[0, \pi]\times [0, 2\pi]\times [0, 5]$ in the example above is just the same as the $[0, 1]^3$ domain from the parametric function defined in Meshes.jl with a rescaling. The parametric function then tells us how we compute the point $\bar r$ from $(\rho, \theta, \phi)$. For the specific parametrization $g:   [0, 5]\times\ [0, 2\pi]\times [0, \pi] \to B_5((0, 0, 0)), (r, \phi, \theta)^T\mapsto (r \sin(\theta)\cdot\cos(\phi), r\sin(\theta)\cdot\sin(\phi), r\cos(\theta))^T$, ($B_5((0, 0, 0))$ being the ball with center $(0, 0, 0)$ and radius 5) we would also have to take the Jacobian determinant of $r^2\sin(\theta)$ into account, while the rescaled parametrization in Meshes.jl has a slightly different determinant of the Jacobian, but essentially does the same.

docs/src/how_it_works.md
```diff
@@ -0,0 +1,77 @@
+# How it Works (By Example)
+
+## Example Problem
+
+Let $f$ be a function of position ($\bar{r}$)
```
JoshuaLampert, line 5: I would maybe use another letter for the position $\bar r\in\mathbb{R}^3$ because IMO something including an $r$ sounds like a radius (especially in the context of a ball). Maybe something like $\boldsymbol{x}$ or maybe better $\boldsymbol{p}$?

docs/src/how_it_works.md
```diff
@@ -0,0 +1,77 @@
...
+\int_0^{\pi} \int_0^{2\pi} \int_0^{5} f(\bar{r}) ~ \text{d}\rho~\text{d}\theta~\text{d}\phi
+```
+
+This works great for simple geometries, but requires integration code that is geometry-specific. This package, by contrast, leverages parametric functions defined in Meshes.jl and differential forms to define integral methods that are general solutions for all geometries.
+
+## Parametric Functions
+
+Every supported `Meshes.Geometry` type is defined as having a parametric function that maps from a local parametric coordinate system to every point on the geometry. Curve-like geometries will have a single parametric dimension, surfaces will have two dimensions, and volumes will have three dimensions; this can be checked for a particular geometry via `Meshes.paramdim(geometry)`.
+
+For consistency across geometry types, these parametric functions are defined to always take coordinates inside a normalized range $[0,1]$. In the example case of `ball`, Meshes.jl defines a parametric function mapped in normalized spherical coordinates $(t_\rho, ~t_\theta, ~t_\phi)$. We find, then:
```
JoshuaLampert, line 36: > these parametric functions are defined to always take coordinates inside a normalized range $[0,1]$

Well, except for `Line` and co. There some of the parametric coordinates can go to infinity in one or both directions.

src/integral.jl
```diff
@@ -3,7 +3,7 @@
 ################################################################################
 
 """
-    integral(f, geometry[, rule]; FP=Float64)
+    integral(f, geometry[, rule]; diff_method=_default_method(geometry,FP), FP=Float64)
```
JoshuaLampert, line 6:
```suggestion
    integral(f, geometry[, rule]; diff_method=_default_method(geometry), FP=Float64)
```

src/integral.jl
```diff
@@ -12,12 +12,13 @@ precision of type `FP`.
...
-- `rule`: optionally, the `IntegrationRule` used for integration (by default `GaussKronrod()` in 1D and `HAdaptiveCubature()` else)
-- `FP`: optionally, the floating point precision desired (`Float64` by default)
+- `rule`: optionally, the `IntegrationRule` used for integration (by default
+`GaussKronrod()` in 1D and `HAdaptiveCubature()` else)
 
-Note that reducing `FP` below `Float64` will incur some loss of precision. By
-contrast, increasing `FP` to e.g. `BigFloat` will typically increase precision
-(at the expense of longer runtimes).
+# Keyword Arguments
+- `diff_method::DifferentiationMethod = _default_method(geometry,FP)`: the method to
```
JoshuaLampert, line 19:
```suggestion
- `diff_method::DifferentiationMethod = _default_method(geometry)`: the method to
```

src/specializations/BezierCurve.jl
```diff
@@ -9,31 +9,43 @@
...
-can be obtained by specifying the use of DeCasteljau's algorithm instead with
-`alg=Meshes.DeCasteljau()` but can come at a steep cost in memory allocations,
-especially for curves with a large number of control points.
+             diff_method=Analytical(), FP=Float64, alg=Meshes.Horner())
+
+Like [`integral`](@ref) but integrates along the domain defined by `curve`.
+
+# Arguments
+- `f`: an integrand function with a method `f(::Meshes.Point)`
+- `curve`: a `BezierCurve` that defines the integration domain
```
JoshuaLampert, line 23:
```suggestion
- `curve`: a `Meshes.BezierCurve` that defines the integration domain
```

JoshuaLampert (changes requested):

docs/src/how_it_works.md
mikeingold: This is all correct. My intention with the quoted statement was simply to convey that, as an end-user presented with this kind of problem, your basic choices are:
- Derive and write all of that code yourself on a per-geometry basis. Then do it again for every new particular geometry.
- Use MeshIntegrals.jl where we've done the work for you, and it's done using a generalized method that "just works".

For context, a big part of my motivation in writing this package came from a prior effort, [JefimenkoModels.jl](https://github.com/mikeingold/JefimenkoModels.jl), a computational EM package that relies on solving time-domain integral equations of current fields over source domains. This was before I even knew about Meshes.jl, so a lot of that work was essentially duplicative effort trying to implement my own Unitful coordinate systems and geometries and mapping them onto integral solvers. When I have to revisit it, I can pretty much just refactor it to depend on Meshes and MeshIntegrals and the only remaining code is generating the particular integrand equations.

docs/src/how_it_works.md
mikeingold: This might be an electromagnetics-specific notation thing, where I'm very used to seeing $\vec{r}$ used to represent an abstract vector from the origin to some observation point, where no particular coordinate system is yet prescribed.

mikeingold: That said, if it’s not universally-understood notation, I can at least clarify the intent in the documentation.

docs/src/how_it_works.md
mikeingold: I just pushed a commit that addresses this. I changed the Triangle-specific page into a Specializations page with a header that outlines the geometries that don't conform to this "standard". I tweaked the language here to caveat "with some notable exceptions" and provide a link to Specializations. Hopefully the new `@ref` links work. (Not a polished solution, but hopefully good enough for today.)

mikeingold: Let me know what you think of the update.

docs/src/specializations.md
```diff
@@ -1,4 +1,18 @@
...
+- `Meshes.FrustumSurface` is essentially a composite type and has a parametric function that only maps the cylindrical portion of the geometry, so the top and bottom `Meshes.Disk` elements have to be integrated separately.
+- `Meshes.Line` represents a line of infinite length that passes through two points, and it has a parametric function that is valid on the domain $(-\infty, \infty)$.
+- `Meshes.Plane` represents a plane of infinite extent, and it has a parametric function that is valid on the domain $(-\infty, \infty)^2$.
+- `Meshes.Ray` represents a line that begins at a point and extends in a particular direction with infinite length, and it has a parametric function that is valid on the domain $[0, \infty)$.
+- `Meshes.Ring` is a composite type that lacks a parametric function, but can be decomposed into `Meshes.Segment`s and then integrated by parts.
+- `Meshes.Rope` is a composite type that lacks a parametric function, but can be decomposed into `Meshes.Segment`s and then integrated by parts.
```
JoshuaLampert, line 11:
```suggestion
- `Meshes.Ring` is a composite type that lacks a parametric function, but can be decomposed into `Meshes.Segment`s and then integrated by adding together the individual integrals.
- `Meshes.Rope` is a composite type that lacks a parametric function, but can be decomposed into `Meshes.Segment`s and then integrated by adding together the individual integrals.
```
Reading "integrated by parts" I immediately think of the integration-by-parts formula, which might be confusing.

docs/src/how_it_works.md
JoshuaLampert: Ok, can we maybe just delete the "by contrast" to avoid confusion?

docs/src/how_it_works.md
JoshuaLampert: Sounds good. I left one suggestion to avoid confusion.

docs/src/how_it_works.md
mikeingold: No problem. Reading it again, I don't think the "by contrast" adds anything to the statement anyway. I just dropped it in a commit.

JoshuaLampert (approved):
Thanks!

PR #123, opened 2024-10-26
Use `@autodocs` to dynamically include all docstrings

JoshuaLampert:
- By using `@autodocs` we don't need to manually list all definitions with docstring. This means that, e.g., in #122 we don't need to change the docs, but the new docstrings are found automatically.
- Added docstrings for spezializations into new section of the API reference. For this I unified the docstrings for `BezierCurve` because otherwise they would appear three times, which would be redundant,
- Added some links and cross references.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,6 +1,12 @@
 using Documenter
 using MeshIntegrals
 
+# Dynamically set all files in subdirectories of the source directory to include all files in these subdirectories.
+# This way they don't need to be listed explicitly.
+SPECIALIZATIONS_FILES = joinpath.(Ref("specializations"),
+    readdir(joinpath(dirname(@__DIR__), "src",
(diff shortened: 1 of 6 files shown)


JoshuaLampert:
What about the specializations? Currently, they are not included in the API list in the docs. I would suggest to add a new subsection `### Specializations` under `## Integrals`, where we dynamically include all files in the folder `specializations`.

JoshuaLampert:
> What about the specializations? Currently, they are not included in the API list in the docs. I would suggest to add a new subsection `### Specializations` under `## Integrals`, where we dynamically include all files in the folder specializations.

I just did that. What do you think, @mikeingold?

mikeingold:
LGTM! Nice work.

Makes progress on #62

mikeingold (approved):

PR #129, opened 2024-11-04
add AD (Enzyme) support via EnzymeExt

kylebeggs:
Adds support to calculate the jacobian via Enzyme autodiff #35 . This is done by adding a package extension for Enzyme, `EnzymeExt`.


Project.toml
```diff
@@ -12,9 +12,16 @@ Meshes = "eacbb407-ea5a-433e-ab97-5258b1ca43fa"
 QuadGK = "1fd47b50-473d-5c70-9696-f719f8f3bcdc"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
+[weakdeps]
+Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
+
+[extensions]
+EnzymeExt = "Enzyme"
```
JoshuaLampert, line 19:
```suggestion
MeshIntegralsEnzymeExt = "Enzyme"
```
I think the naming convention for extensions is to also include the name of the base package. Of course this also requires to update the filename `EnzymeExt.jl` to `MeshIntegralsEnzymeExt.jl` and the name of the module.

ext/EnzymeExt.jl
```diff
@@ -0,0 +1,12 @@
+module EnzymeExt
+
+using MeshIntegrals
+using Meshes: Geometry
+using Enzyme
+
+function MeshIntegrals.jacobian(
+        geometry::Geometry, ts::V, ::AutoEnzyme) where {V <: Union{AbstractVector, Tuple}}
+    to.(Enzyme.jacobian(Enzyme.Forward, geometry, ts...))
```
JoshuaLampert, line 9: Would it make sense to include `Enzyme.Forward` as an option to `AutoEnzyme`, such that one could also use `Enzyme.Backward` mode?

ext/EnzymeExt.jl
```diff
@@ -0,0 +1,12 @@
+module EnzymeExt
+
+using MeshIntegrals
+using Meshes: Geometry
+using Enzyme
```
JoshuaLampert, line 5:
```suggestion
using Enzyme: Enzyme
```
We generally prefer explicit imports.

ext/EnzymeExt.jl
```diff
@@ -0,0 +1,12 @@
+module EnzymeExt
+
+using MeshIntegrals
```
JoshuaLampert, line 3:
```suggestion
using MeshIntegrals: MeshIntegrals, AutoEnzyme
```

ext/EnzymeExt.jl
```diff
@@ -0,0 +1,12 @@
...
+using Meshes: Geometry
```
JoshuaLampert, line 4:
```suggestion
using Meshes: Geometry, to
```
I see you imported `to` in the main MeshIntegrals.jl file. Would it make sense to only import it in the extension?

JoshuaLampert (changes requested):
This looks pretty nice already. Great to see that Enzyme works pretty much out of the box. I only have some minor comments/suggestions.

ext/EnzymeExt.jl
```diff
@@ -0,0 +1,12 @@
+module EnzymeExt
```
JoshuaLampert, line 1:
```suggestion
module MeshIntegralsEnzymeExt
```
See above.

src/MeshIntegrals.jl
```diff
@@ -1,7 +1,7 @@
 module MeshIntegrals
 using CliffordNumbers: CliffordNumbers, VGA, ∧
 using CoordRefSystems: CoordRefSystems, CRS
-using Meshes: Meshes, Geometry
+using Meshes: Meshes, Geometry, to
```
JoshuaLampert, line 4:
```suggestion
using Meshes: Meshes, Geometry
```
This would not be necessary anymore with the suggestion above.

JoshuaLampert (changes requested):

ext/EnzymeExt.jl
kylebeggs: This is a good point. If one did a reverse mode calc on a higher-level function such as `integral`, would this be an issue or would Enzyme just take care of it? I need to think this through. Ideas?

mikeingold:
@kylebeggs is there any chance you’re better at `git rebase` than I am? With the recent merged PR’s it’s made the change summary a little challenging to parse.

kylebeggs:
I don't think anyone is actually good at git rebase haha. I'll give it a shot. Sometimes I find it easier to just make a new PR when there the current PR changes aren't too sparse like this one. I'm traveling this weekend so it'll be Monday.

mikeingold:
> I don't think anyone is actually good at git rebase haha. I'll give it a shot. Sometimes I find it easier to just make a new PR when there the current PR changes aren't too sparse like this one. I'm traveling this weekend so it'll be Monday.

It always sounds easy enough in practice but fails miserably for me. I just went through the same with #114 which became #130 lol. No worries on timeline. I think the additions here look good.

kylebeggs:
@mikeingold I will still work on this in the sense that a few of the tests error and need to address that, but will wait until #131 is done to not further complicate the rebasing.

ext/MeshIntegralsEnzymeExt.jl
mikeingold, line 1: I just pushed a commit that tweaked formatting and updated the argument parameterizations to the latest style, based on how things changed with #122.

ext/MeshIntegralsEnzymeExt.jl
```diff
@@ -0,0 +1,15 @@
...
+using MeshIntegrals: MeshIntegrals, AutoEnzyme
+using Meshes: Geometry, to
+using Enzyme: Enzyme
+
+function Meshes.jacobian(
+        geometry::Geometry,
+        ts::Union{AbstractVector{T}, Tuple{T, Vararg{T}}},
+        ::AutoEnzyme
+) where {T <: AbstractFloat}
+    return to.(Enzyme.jacobian(Enzyme.Forward, geometry, ts...))
```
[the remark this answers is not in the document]
kylebeggs: I haven't tested it, but I would be surprised if not

kylebeggs:
@mikeingold @JoshuaLampert Any ideas on how to go about adding the tests? Maybe wrpa the current tests in `foreach(ad_backends) do backend`? Is there a way to run the test set with twice with different setups? That would be the cleanest, but I can't figure that out.

mikeingold:
tl;dr-up-front: I'm not sure what the best answer is. 

I don't think we necessarily need exhaustive coverage of the entire combination space of: integrand function output types, geometry types, integration rule type and settings, differentiation method, floating point precision spec, etc. It would be nice, but could lead to a combinatorial explosion at test time.

The `DifferentiationMethod`s are really just used to determine `differential` values via `jacobian` (and since `jacobian` is currently exported), so it might be enough to just write tests that check those results directly. We could also just add an extra test to each `combinations.jl` test set for an integral using it.

JoshuaLampert:
Another option would be to have separate jobs for the different backends. Meshes.jl does something similar by setting an environment variable, see [here](https://github.com/JuliaGeometry/Meshes.jl/blob/7e94faf23ba8c05e3ef4642ab8dd4132653d3f05/test/runtests.jl#L34-L42) and [here](https://github.com/JuliaGeometry/Meshes.jl/blob/7e94faf23ba8c05e3ef4642ab8dd4132653d3f05/.github/workflows/CI.yml#L33-L35) and [here](https://github.com/JuliaGeometry/Meshes.jl/blob/7e94faf23ba8c05e3ef4642ab8dd4132653d3f05/.github/workflows/CI.yml#L55). This would save us some CI time and the different jobs can fail and succeed independently of each other making it easier to investigate where a problem is if a job fails.

kylebeggs:
@mikeingold I think that is a sufficient approach. @JoshuaLampert I initially tried to go this route, but I realized there is no good way to run the test suite locally in that manner. I guess you can set the variable and then run the macro that runs all the tests in the REPL?

JoshuaLampert:
Honestly, I think the easiest would be to just change the analogue of [this line](https://github.com/JuliaGeometry/Meshes.jl/blob/7e94faf23ba8c05e3ef4642ab8dd4132653d3f05/test/runtests.jl#L41) locally from `Float64` to `Float32` (or in our case from `FiniteDifference()` to `AutoEnzyme()`.

kylebeggs:
I decided to try the route of full integration tests. If we decide the additional tests will extend the total test time too much, we can just test `jacobian`, but I think this would be the ideal approach. Most of the tests pass, but there are 4 or 5 that error due to Enzyme such as `Cylinder`, so I guess I need to write custom rules for those.

test/runtests.jl
```diff
@@ -8,4 +8,34 @@ using TestItems
...
     using Meshes
     using Unitful
+
+    # environment settings
+    isCI = "CI" ∈ keys(ENV)
+
+    # float settings
+    AD = if isCI
+        if ENV["AD"] == "FiniteDifference"
+            FiniteDifference()
```
mikeingold, line 18: Currently, one snag with this approach is that several of the geometries with `specialization` methods are edge cases that use `integral` methods with hard-coded `Analytical` differentials. Trying to integrate those geometries with any other `DifferentiationMethod` is guaranteed to throw an error.

This was one of the reasons I was hoping to just eliminate `Analytical` altogether. They're faster for a couple of specific geometry types, but at the cost of a fair bit of added complexity. There was some benefit in code/math reduction, but largely it was to eliminate these cases where `diff_method` isn't truly supported.

mikeingold: I'm planning to write up a Discussion post with my thoughts, and some historical context, on our unit test structure/organization. I suspect it might be adequate to just add a single new test to each geometry's standard test set, something like
```julia
@test integral(f, geometry, GaussLegendre(100); diff_method=AutoEnzyme())
```

I think the impact of this new `diff_method` should be apparent for any integral of each geometry type, which would make re-running all of them redundant. And, especially with other recent performance and CI runtime improvements, we currently have a little CI runtime to spare.

kylebeggs: > Currently, one snag with this approach is that several of the geometries with specialization methods are edge cases that use integral methods with hard-coded Analytical differentials. Trying to integrate those geometries with any other DifferentiationMethod is guaranteed to throw an error.

I left those tests to use `integral`, not the `integral_test` functions I created in `Setup`, this is not an issue.

> This was one of the reasons I was hoping to just eliminate Analytical altogether. They're faster for a couple of specific geometry types, but at the cost of a fair bit of added complexity. There was some benefit in code/math reduction, but largely it was to eliminate these cases where diff_method isn't truly supported.

Personally, I think the added complexity is worth the speed. But that is just my opinion.

> I suspect it might be adequate to just add a single new test to each geometry's standard test set, something like

This makes sense. I will change to this approach.

PR #135, opened 2024-11-25
Implement _zeros and _ones utils

mikeingold:
## Changes
- Implements `_zeros` and `_ones` utility functions. These act like `Base.zeros` and `Base.ones`, respectively, but return `NTuple`s to eliminate some allocations.
  - Improves performance for `HAdaptiveCubature` integrals, up to about 2.7x for particularly simple integrals.

## Notes
- Originally included in #131, but partitioned for scope management.

### DIFF
--- a/src/specializations/Plane.jl
+++ b/src/specializations/Plane.jl
@@ -91,7 +91,7 @@ function integral(
     # Create a wrapper that returns only the value component in those units
     uintegrand(uv) = Unitful.ustrip.(integrandunits, integrand(uv))
     # Integrate only the unitless values
-    value = HCubature.hcubature(uintegrand, -ones(FP, 2), ones(FP, 2); rule.kwargs...)[1]
+    value = HCubature.hcubature(uintegrand, -_ones(FP, 2), _ones(FP, 2); rule.kwargs...)[1]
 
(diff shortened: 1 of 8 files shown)


src/specializations/Plane.jl
```diff
@@ -91,7 +91,9 @@ function integral(
     # Create a wrapper that returns only the value component in those units
     uintegrand(uv) = Unitful.ustrip.(integrandunits, integrand(uv))
     # Integrate only the unitless values
-    value = HCubature.hcubature(uintegrand, -ones(FP, 2), ones(FP, 2); rule.kwargs...)[1]
+    a = 0 .- _ones(FP, 2)
```
JoshuaLampert, line 94:
```suggestion
    a = .- _ones(FP, 2)
```
I think you don't need the zero and it looks a bit odd to me.

JoshuaLampert (commented):
Looks good. I only have one minor comment.

src/specializations/Plane.jl
mikeingold: Didn’t realize that worked.

JoshuaLampert (approved):
Thanks!

PR #136, opened 2024-11-27
Deprecate applying Gauss-Kronrod rules to surfaces

mikeingold:
## Changes
- Deprecates support for integrating (non-specialization) surfaces using `GaussKronrod` rules, issuing a `Base.depwarn` deprecation warning.
- Eliminates the `_integral_gk_*d` helper functions - consolidated into `_integral(f, g, ::GaussKronrod)` function instead.
- Minor tweaks
  - Removed explicit `FP` conversions inside `GaussLegendre` method, relying instead on automatic promotion from rational to float, resulting in a >2x performance improvement.
  - Removed obsolete `QuadGK` direct dependency from tests.

## Discussion

Surfaces are geometries where `paramdim(geometry) == 2`. Use of nested `GaussKronrod` rules for these geometries is strictly inferior to simply using an equivalent `HAdaptiveCubature` rule, where generally returns results faster and with more accuracy. Earlier in the development of this package, using `GaussKronrod` rules on surfaces provided a workaround since `HAdaptiveCubature` rules didn't yet support Unitful-valued integrand functions. Since that issue has been resolved, I think we should mark this usage as deprecated in `v0.16` and plan to remove support in `v0.17`. Support for usage on several of the specialization geometries would be unaffected for now, where there remains some justification for their usage.


mikeingold:
@JoshuaLampert can you sanity check the `_integral(f, geometry, ::GaussKronrod)` function for me? For some reason the `integrand` symbol isn't being recognized, but I just don't see the issue. I tried commenting out the preceding `depwarn` in case there were any weird side effects, but saw no change.

docs/src/supportmatrix.md
```diff
@@ -1,58 +1,61 @@
...
 
-While this library aims to support all possible integration rules and [**Meshes.jl**](https://github.com/JuliaGeometry/Meshes.jl)
-geometry types, some combinations are ill-suited and some others are simply not yet
-implemented. The following Support Matrix aims to capture the current development state of
-all geometry/rule combinations. Entries with a green check mark are fully supported
-and have passing unit tests that provide some confidence they produce accurate results.
+This library aims to enable users to calculate the value of integrals over all [**Meshes.jl**](https://github.com/JuliaGeometry/Meshes.jl)
+geometry types using an array of numerical integration rules and techniques. However, some
+combinations of geometry types and integration rules are ill-suited, and some others are simply
+not yet yet implemented. The following Support Matrix captures the current state of support for
```
JoshuaLampert, line 6:
```suggestion
not yet implemented. The following Support Matrix captures the current state of support for
```

docs/src/supportmatrix.md
```diff
@@ -1,58 +1,61 @@
...
+unit tests designed to check for accuracy.
 
 In general, Gauss-Kronrod integration rules are recommended (and the default) for geometries
-with one parametric dimension, e.g.: `Segment`, `BezierCurve`, and `Rope`. Gauss-Kronrod
-rules can also be applied to some geometries with more dimensions by nesting multiple
-integration solves, but this is inefficient. These Gauss-Kronrod rules are supported (but
-not recommended) for surface-like geometries, but not for volume-like geometries. For
-geometries with more than one parametric dimension, e.g. surfaces and volumes, H-Adaptive
-Cubature integration rules are recommended (and the default).
+with one parametric dimension, e.g.: `Segment`, `BezierCurve`, and `Rope`. or geometries with
```
JoshuaLampert, line 11:
```suggestion
with one parametric dimension, e.g.: `Segment`, `BezierCurve`, and `Rope`. For geometries with
```

JoshuaLampert (changes requested):
Looks mostly good to me. Please find some comments below.

src/integral.jl
```diff
@@ -33,22 +33,31 @@ function integral(
...
         rule::GaussKronrod;
-        kwargs...
-)
-    # Pass through to dim-specific workers in next section of this file
-    N = Meshes.paramdim(geometry)
+        FP::Type{T} = Float64,
+        diff_method::DM = _default_method(geometry)
+) where {DM <: DifferentiationMethod, T <: AbstractFloat}
+    # Implementation depends on number of parametric dimensions over which to integrate
+    const N = Meshes.paramdim(geometry)
```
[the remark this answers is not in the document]
mikeingold: Interesting. I knew that it was recommended in global scope but didn’t realize it was forbidden in local scope.

mikeingold:
Thanks!

mikeingold:
Good to go here. Looks like the tweaks to `GaussLegendre` methods achieved a surprisingly big performance improvement.

JoshuaLampert (approved):
> Good to go here. Looks like the tweaks to `GaussLegendre` methods achieved a surprisingly big performance improvement.

That's nice! Thanks!

PR #137, opened 2024-11-28
Generalize integrand `f` to any callable

### LINKED ISSUE #132: Generalize integrand `f` to any callable
For some `integral(f, geometry...)`, our API currently requires `f::Function`. However, the only real requirement for `f` internally is that it must expose a method `f(::Point)`, which can be fulfilled by other callables. For example:
```julia
struct Widget
    ...
end

(w::Widget)(p::Point) = ...

w = Widget(...)
geometry = ...

integral(w, geometry)
```

Enabling this would require
- a sweep of all `integral` and `_integral` methods to remove type params from `f`
- updating documentation, e.g. mentioning this fact in the README
- evaluating whether this causes any dispatch issues
- adding a test that specifies a callable non-`Function` for `f`

mikeingold:
Will close #132

### DIFF
--- a/src/integral.jl
+++ b/src/integral.jl
@@ -24,7 +24,7 @@ function integral end
 
 # If only f and geometry are specified, select default rule
 function integral(
-        f::Function,
+        f,
         geometry::Geometry,
         rule::I = Meshes.paramdim(geometry) == 1 ? GaussKronrod() : HAdaptiveCubature();
         kwargs...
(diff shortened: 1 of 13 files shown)


mikeingold:
The codecov/patch test fails because these changes touch the Tetrahedron methods, whose tests are currently tagged under extended/local-only testing.

JoshuaLampert (approved):
LGTM. Thanks!

PR #139, opened 2024-11-30
WIP: Implement _ParametricGeometry for Line

mikeingold:


mikeingold:
Thanks @JoshuaLampert for all the support lately. My free dev time seems to come in waves lol.

mikeingold:
Just noticed that I didn't update the `_parametric` methods for `Triangle` and `Tetrahedron`, which don't fit nicely into the new methodology. I'll have to think about those for a minute.

src/specializations/Tetrahedron.jl
```diff
@@ -27,17 +26,20 @@ end
...
 
-function _parametric(tetrahedron::Meshes.Tetrahedron, t1, t2, t3)
-    if (t3 < 0) || (t3 > 1)
-        msg = "tetrahedron(t1, t2, t3) is not defined for t3 outside [0, 1]."
-        throw(DomainError(t3, msg))
-    end
+function _parametric(tetrahedron::Meshes.Tetrahedron)
+    function f(t1, t2, t3)
+        if any(Iterators.map(n -> (n < 0) || (n > 1), (t1, t2, t3)))
+            msg = "tetrahedron(t1, t2, t3) is not defined for t outside [0, 1]."
```
JoshuaLampert, line 32:
```suggestion
            msg = "tetrahedron(t1, t2, t3) is not defined for (t1, t2, t3) outside [0, 1]³."
```

src/specializations/Triangle.jl
```diff
@@ -111,16 +111,19 @@ _has_analytical(::Type{T}) where {T <: Meshes.Triangle} = true
...
 ################################################################################
 
-function _parametric(triangle::Meshes.Triangle, t1, t2)
-    if (t1 < 0 || t1 > 1) || (t2 < 0 || t2 > 1)
-        msg = "triangle(t1, t2) is not defined for (t1, t2) outside [0, 1]^2."
-        throw(DomainError((t1, t2), msg))
+function _parametric(triangle::Meshes.Triangle)
+    function f(t1, t2)
+        if any(Iterators.map(n -> (n < 0) || (n > 1), (t1, t2)))
+            msg = "triangle(t1, t2) is not defined for (t1, t2) outside [0, 1]^2."
```
JoshuaLampert, line 117:
```suggestion
            msg = "triangle(t1, t2) is not defined for (t1, t2) outside [0, 1]²."
```
This just looks a bit nicer (and is how it looks like in Meshes.jl) :smiley:

src/specializations/Ray.jl
```diff
@@ -11,85 +11,24 @@
...
 
 ################################################################################
-#                               jacobian
+#                                 Parametric
 ################################################################################
 
-_has_analytical(::Type{T}) where {T <: Meshes.Ray} = true
+# Map argument domain from [0, 1] to [0, ∞) for (::Ray)(t)
+# f(t) = t / 1 - t^2)
+# f'(t) = (t^2 + 1) / (1 - t^2)^2
```
JoshuaLampert, line 30:
```suggestion
# f(t) = t / (1 - t^2)
```

JoshuaLampert (changes requested):
Looks good to me in general. I just have some small suggestions. I do have one question:
So `Triangle` is the only geometry, which uses a guaranteed analytical derivative? Do I understand correctly that you propose to use `_ParametricGeometry` for `Line`, `Plane`, and `Ray` to simplify the code despite it being slower, while for `Triangle` the performance loss is too high by using `_ParametricGeometry`, which is why we still use this special method for the integration?

JoshuaLampert:
Some fun bikeshedding: Interestingly, in the newest version the spell checker doesn't like "parametrized" and wants it to be spelled as "parameterized". As a non-native English speaker I'm not sure, but always assumed both versions are fine (seeing you usually use "parameterized"). [Wikipedia](https://en.wikipedia.org/wiki/Parametrization_(geometry)) also says both are fine. As `ParametrizedCurve` is defined upstream anyway, the best solution is probably to just put that into a `.typos.toml` file.

mikeingold:
> Some fun bikeshedding: Interestingly, the newest version of the spell checker doesn't like "parametrized" and wants it to be spelled as "parameterized". As a non-native English speaker I'm not sure, but always assumed both versions are fine (seeing you usually use "parameterized"). [Wikipedia](https://en.wikipedia.org/wiki/Parametrization_(geometry)) also says both are fine. As `ParametrizedCurve` is defined upstream anyway, the best solution is probably to just put that into a `.typos.toml` file. Edit: Just did that and the spell checker is happy again.

I actually hadn’t even realized the Meshes geometry was spelled that way until I saw the tagged bot comments last night. I often find that compound technical words, especially when they’re domain-specific, are flagged as mis-spellings by common spell checkers, so I think I’ve developed an unfortunate tendency to ignore those warnings.

Looks like both spellings, with and without the “e” are valid but each is maybe more commonly used in certain contexts. If one is more widely preferred then I’d have no reservations about adjusting all of the spellings in this package.

mikeingold:
> Looks good to me in general. I just have some small suggestions. I do have one question: So `Triangle` is the only geometry, which uses a guaranteed analytical derivative? Do I understand correctly that you propose to use `_ParametricGeometry` for `Line`, `Plane`, and `Ray` to simplify the code despite it being slower, while for `Triangle` the performance loss is too high by using `_ParametricGeometry`, which is why we still use this special method for the integration?

My personal preference would be to just “rip off the bandage” and implement them all, then focus on restoring some of the performance loss in patch updates, but I don’t want to dictate the path we take if that isn’t palatable to others.

I went ahead and implemented these because the performance loss was less impactful, and I’m more willing to argue for these.

JoshuaLampert:
I would be fine with the performance overhead for `Line`, `Plane`, and `Ray` _if_ it really is "only" a factor of not more than two as the benchmark results in this PR suggest. I would be more hesitant if the factor would be something like 20 as it was in the former PR for `Plane` (see https://github.com/JuliaGeometry/MeshIntegrals.jl/pull/131#issuecomment-2467173076). Do you know why it was so much slower in the other PR, @mikeingold? I would also like to hear @kylebeggs' opinion since [he expressed some doubts in the first PR](https://github.com/JuliaGeometry/MeshIntegrals.jl/pull/131#issuecomment-2476353978).

mikeingold:
> I would be fine with the performance overhead for `Line`, `Plane`, and `Ray` _if_ it really is "only" a factor of not more than two as the benchmark results in this PR suggest.

To me it’s some qualitative combination of: relative performance difference, absolute timescale, and how we expect this geometry to be used (Triangle is a simplex, so more likely to be performance sensitive if we want to integrate something like a surface mesh).

> I would be more hesitant if the factor would be something like 20 as it was in the former PR for `Plane` (see [#131 (comment)](https://github.com/JuliaGeometry/MeshIntegrals.jl/pull/131#issuecomment-2467173076)). Do you know why it was so much slower in the other PR, @mikeingold?

I actually hadn’t caught that. For some reason I remembered Triangle being about 5x slower and didn’t recall the old Plane benchmark being so bad. It looks like the reference times are both about 500 us, so I’m not sure what’s so different about the two implementations. The switch to `_parametric` yielding a function came after and performance looked unchanged, so it’s not that. I’ll have to compare the code to see what’s different, because it was mostly a copy/paste from the old `tetrahedron` branch.

mikeingold:
The only obvious relevant difference in seeing between the two branches is the dropped `FP` conversions in the general `GaussLegendre` function since last time. I wonder if the is simply that the general case has gotten more performant and closed the gap.

src/specializations/Triangle.jl
mikeingold: Agreed. I just commit other changes to this file and included the suggested change there.

mikeingold:
I just tried out the `_ParametricGeometry` form for `Triangle` again. The baseline reference times were basically identical. Performance improved significantly vs the last attempt in #131, but it's still a significant performance penalty.

PR | `main` time | PR time | Degradation
---|---|---|---
#131 | 0.68 ms | 34 ms | 50x
This one | 0.70 ms | 20 ms | 27x

I'm willing to argue that a regression from 0.5 ms to 0.9 ms for integrating an infinite plane is acceptable given the other benefits, but this is a lot more impactful, especially given its status as the 2D simplex. I'll see if I can do some profiling to improve performance, but without significant improvements I'm going to have to revert this change and defer it again.

mikeingold:
I think I pulled it off. I refactored `_parametric(::Triangle)` to use a more clever formulation. The old version needed two calls to `(::Meshes.Triangle)(t1, t2)`, formed a `Segment` between those points, and then a call to `(::Meshes.Segment)(t)`. The new version calculates the equivalent barycentric coordinates directly and then only needs a single `(::Meshes.Triangle)(t1, t2)` call.

We've gone from a 27x performance degradation to only about 10%! 🎉 As a bonus, performance for integrals of `Tetrahedron` is also improved by about 10x since it relies on the same function under the hood.

src/specializations/Ray.jl
mikeingold: I left `f'(t)` as a hint to future-me in case I wanted to try implementing an analytical derivative. That's no longer the case, so I went ahead and removed both lines.

mikeingold:
> Some fun bikeshedding: Interestingly, the newest version of the spell checker doesn't like "parametrized" and wants it to be spelled as "parameterized". As a non-native English speaker I'm not sure, but always assumed both versions are fine (seeing you usually use "parameterized"). [Wikipedia](https://en.wikipedia.org/wiki/Parametrization_(geometry)) also says both are fine. As `ParametrizedCurve` is defined upstream anyway, the best solution is probably to just put that into a `.typos.toml` file. Edit: Just did that and the spell checker is happy again.

I just added an Issue #143 to keep track of this point.

I'm pretty happy with where this PR is at, despite it growing in scope from what I'd originally planned.

src/specializations/Triangle.jl
```diff
@@ -12,115 +12,29 @@
...
-    # Integrate only the unitless values
-    ∫ = HCubature.hcubature(uintegrand, _zeros(FP, 2), (T(1), T(π / 2)), rule.kwargs...)[1]
-
-    # Apply a linear domain-correction factor 0.5 ↦ area(triangle)
-    return 2 * Meshes.area(triangle) * integrandunits .* ∫
+        rule::IntegrationRule;
+        kwargs...
+)
+    # Generate a _ParametricGeometry whose parametric function domain spans [0,1]²
+    param_triangle = _ParametricGeometry(_parametric(triangle), 2)
```
JoshuaLampert, line 19:
```suggestion
    param_triangle = _ParametricGeometry(_parametric(triangle), Meshes.paramdim(triangle))
```

src/specializations/Triangle.jl
```diff
@@ -12,115 +12,29 @@
...
+# Map argument domain from [0, 1]² to Barycentric domain for (::Triangle)(t1, t2)
+function _parametric(triangle::Meshes.Triangle)
+    function f(t1, t2)
+        if any(Iterators.map(n -> (n < 0) || (n > 1), (t1, t2)))
+            msg = "triangle(t1, t2) is not defined for (t1, t2) outside [0, 1]²."
+            throw(DomainError((t1, t2), msg))
+        end
 
-    return segment(t1)
+        return triangle(t1 * t2, t2 - (t1 * t2))
```
JoshuaLampert, line 37:
```suggestion
        t1t2 = t1 * t2
        return triangle(t1t2, t2 - t1t2)
```
I don't know how much we gain from this, but since this function is performance critical maybe saving one multiplication in each call gives us a small performance gain? We could at least try.

src/specializations/Tetrahedron.jl
```diff
@@ -27,17 +26,21 @@ end
...
-    c = tetrahedron(0, 0, t3)
-    cross_section = Meshes.Triangle(a, b, c)
+        # Take a triangular cross-section at t3
+        a = tetrahedron(t3, 0, 0)
+        b = tetrahedron(0, t3, 0)
+        c = tetrahedron(0, 0, t3)
+        cross_section = _parametric(Meshes.Triangle(a, b, c))
 
-    return _parametric(cross_section, t1, t2)
+        return cross_section(t1, t2)
```
JoshuaLampert, line 43: [IIUC correctly](https://github.com/JuliaGeometry/MeshIntegrals.jl/pull/139#issuecomment-2510437666) you managed to obtain a significant speedup for `Triangle` by calculating the barycentric coordinates directly from `t1`, `t2`, `t3`. Could we apply a similar trick also for `tetrahedron`, such that we only need to call `tetrahedron` once?

JoshuaLampert (changes requested):
This already looks pretty good. I would guess that the main reason for the performance decrease is the fact that we need to apply finite differences now. Therefore, my hope is that we will gain some performance again after we have finished the `AutoEnzyme` backend.
Can you please remove the reference to `Analytical` in https://github.com/JuliaGeometry/MeshIntegrals.jl/blob/c8f66c545cfed1b57e6d3dc09d4ab9f831fc85ff/src/differentiation.jl#L12 to make the Documenter happy? (I cannot comment on that line because it is not part of the diff.)

kylebeggs:
@JoshuaLampert  My main reason for concern is precisely what @mikeingold stated:

> Triangle is a simplex, so more likely to be performance sensitive if we want to integrate something like a surface mesh

but 10% is not too bad at all, and hopefully using autodiff does help.

The code simplification is enormous here, it looks great!

src/specializations/Tetrahedron.jl
mikeingold: The update to `_parametric(::Triangle)` was possible because we only needed to locate a `Point` on a given `Triangle`. Something similar is likely possible for `_parametric(::Tetrahedron)` but would take some time to derive. The simplest path was to locate a triangular cross-section via `t3` and then use `(t1, t2)` as triangle coords, but for optimal performance we'd need to just directly derive 3D-barycentric coords and return something of the form `tetrahedron(a, b, c)`. Definitely worth pursuing, so I documented it in #145 for hopefully near-future work.

JoshuaLampert: Nice, thanks! Definitely fine to leave it as is for now.

JoshuaLampert (approved):
Thanks!

PR #144, opened 2024-12-02
Use default `arch` in CI

JoshuaLampert:
There are warnings (e.g., https://github.com/JuliaGeometry/MeshIntegrals.jl/actions/runs/12112016118?pr=139) in the CI runs because we request x64 on a macOS runner with arm64. I propose to simply use the default architecture, i.e. to remove the arch argument. For the flag-name of the coveralls action I followed the example under [coverallsapp/github-action#complete-parallel-job-example](https://github.com/coverallsapp/github-action?tab=readme-ov-file#complete-parallel-job-example).

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -30,7 +30,7 @@ concurrency:
 jobs:
   test:
     if: "!contains(github.event.head_commit.message, 'skip ci')"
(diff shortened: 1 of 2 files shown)


mikeingold:
LGTM

mikeingold (approved):

Issue #146, opened 2024-12-02
Changes to `_default_method`


mikeingold:
## Background

The `_default_method` utility function currently takes a `(::Geometry)` or `(::Type{<:Geometry})` as argument since it depended on the geometry being integrated. PR #139 potentially makes this obsolete by removing `Analytical` and making `FiniteDifference` the universal default.

## Ideas

I think the function should be renamed `_default_diff_method` for clarity.

Should this function continue to take geometry-information as an argument? We're not currently using that information, but it's possible that could change again in the future as more `DifferentiationMethod`s are implemented.

JoshuaLampert:
I think it doesn't harm if `_default_method` takes a `geometry`. It shouldn't decrease performance, but allows for more flexibility in the future. I agree that renaming it to `_default_method` makes sense.

PR #147, opened 2024-12-02
Rename _default_method and absorb _gausslegendre

mikeingold:
## Changes
- Renames `_default_method` to `_default_diff_method` (#146)
- Absorbed `_gausslegendre` into only remaining `GaussLegendre` integral method
  - This change improved performance up to about 30% for `GaussLegendre` integrals.
- Reorganized `utils.jl` functions and added docstrings


JoshuaLampert (approved):
Thanks.

PR #149, opened 2024-12-04
Calculate nodes and weights at `GaussLegendre` construction time

### LINKED ISSUE #148: Possible to cache Gauss-Legendre nodes and weights
When integrating using a `GaussLegendre` rule, we calculate nodes and weights using `FastGaussQuadrature.gausslegendre(::Int) -> Tuple{Vector, Vector}`.

For geometries that need to be decomposed prior to integration, e.g. `Rope -> Segment`, the same nodes and weights will be re-calculated for each component of the domain. It might be beneficial, instead, to calculate them once and then pass them through as an optional or keyword argument to `_integral`.

mikeingold:
## Changes
- Adds `nodes` and `weights` as stored fields to `GaussLegendre <: IntegrationRule`. These are generated at instantiation-time instead of at `integral`-time so that the results can be re-used.
  - Closes #148 

## Discussion

The Gauss-Legendre nodes and weights are purely a function of order `n`, so I think it makes sense to just calculate them when constructing the rule itself. This will lead to some moderate performance benefits for rules used more than once, or where the integration occurs in some performance-sensitive user code block.

Currently, FastGaussQuadrature.jl doesn't allow us to request nodes and weights in a particular `eltype`, and instead always returns `Vector{Float64}`s. If that support ever materializes then I think it could make sense to add a parameter `GaussLegendre{T}` to handle this, but for now I don't think it's necessary.

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -45,6 +45,7 @@ spec = (
         line = Line(Point(0, 0, 0), Point(1, 1, 1)),
         plane = Plane(Point(0, 0, 0), Vec(0, 0, 1)),
         ray = Ray(Point(0, 0, 0), Vec(0, 0, 1)),
+        rope = Rope([Point(t, t, t) for t in 1:32]...),
         triangle = Triangle(Point(1, 0, 0), Point(0, 1, 0), Point(0, 0, 1)),
         tetrahedron = let
             a = Point(0, 3, 0)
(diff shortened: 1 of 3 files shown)


JoshuaLampert (approved):
Nice! This can also be helpful if you want to integrate multiple functions even in potentially different geometries because you can reuse the rule object, i.e. also in this case you only need to compute the nodes and weight once.

PR #151, opened 2024-12-04
Standardization of combinations tests

mikeingold:
## Changes
- Standardized `combinations` tests by wrapping testable items in a data structure and passing them to a test suite.
  - Reduces boilerplate and subtle variations between `@testitem`s.
  - Added Unitful integrands to several tests that were using simple scalars.
  - Added commented-out test infrastructure for upcoming `AutoEnzyme` in #129.
  - Re-organized to have `@testsnippet`s hosted in the source file where they're used.


test/combinations.jl
```diff
@@ -1,429 +1,365 @@
...
+
+    function runtests(
+            testable::TestableGeometry,
+            supports::SupportStatus;
+            rtol = sqrt(eps())
+    )
+        # Test alias functions
+        for alias in (lineintegral, surfaceintegral, volumeintegral)
+            # if supports.alias
+            if getfield(supports, first(methods(alias)).name)
```
JoshuaLampert, line 76:
```suggestion
            if getfield(supports, nameof(alias))
```
I think this would be the more natural way to obtain the name of the function as a symbol.

test/combinations.jl
```diff
@@ -1,429 +1,365 @@
...
-
-    # Integral aliases
-    @test_throws "not supported" lineintegral(f, ball)
-    @test surfaceintegral(f, ball) ≈ sol
-    @test_throws "not supported" volumeintegral(f, ball)
+    solution = (π - π * exp(-radius^2)) * u"A*m^2"
+
+    # Package and run tests
+    testable = TestableGeometry(integrand, ball, solution)
+    runtests(testable, SupportStatus(:surface))
```
JoshuaLampert, line 142:
```suggestion
    runtests(testable, SupportStatus(ball))
```
If we want to apply my suggestion from above. This would need to be done for each geometry.

test/combinations.jl
```diff
@@ -1,429 +1,365 @@
...
+- The appropriate alias function, e.g. `lineintegral` for 1D geometries, works.
+- Invalid applications of integral aliases produce a descriptive and useful error message.
+- (Planned) Usage of non-default DifferentiationMethods.
+"""
+
+#===============================================================================
+                        Test Generation Infrastructure
+===============================================================================#
+
+@testsnippet Combinations begin
```
JoshuaLampert, line 16: Does this mean, we can delete the `Setup` `@testsnippet` in runtests.jl?

test/combinations.jl
```diff
@@ -436,477 +372,296 @@ end
...
-        @test_throws "not supported" volumeintegral(f, curve_cart)
-        @test lineintegral(f, curve_polar) ≈ sol
-        @test_throws "not supported" surfaceintegral(f, curve_polar)
-        @test_throws "not supported" volumeintegral(f, curve_polar)
+        solution = 2π * radius * exp(-radius^2) * u"A*m"
+
+        # Package and run tests
+        testable_cart = TestableGeometry(integrand, curve_cart, solution)
+        runtests(testable_cart, SupportStatus(:line))
+        testable_polar = TestableGeometry(integrand, curve_cart, solution)
```
JoshuaLampert, line 461:
```suggestion
        testable_polar = TestableGeometry(integrand, curve_polar, solution)
```

test/combinations.jl
```diff
@@ -1,429 +1,365 @@
...
+            rules = Bool.((1, 1, 1))
+            return SupportStatus(aliases..., rules...)
+        elseif sym == :volume
+            aliases = Bool.((0, 0, 1))
+            rules = Bool.((0, 1, 1))
+            return SupportStatus(aliases..., rules...)
+        else
+            error("Unrecognized SupportStatus shortcut $(string(sym))")
+        end
+    end
```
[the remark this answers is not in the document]
mikeingold: I think this is a good idea. When I first implemented this I expected there to be more outliers that didn’t conform to the standard support structure, but after the recent `_ParametricGeometry` PR’s it looks like support is currently purely a function of `paramdim`s, which is really nice. I’ll try to get all of the tests converted when I get back to my computer.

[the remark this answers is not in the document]
mikeingold: I forgot to go back and add a question about Box-4D. It’s currently on extended/local-only testing because it’s niche and CI times are pretty lengthy. In terms of implementation for `Box` methods, there’s actually nothing special about 2D vs 4D vs 32D, etc. Should we bother keeping that test set?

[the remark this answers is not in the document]
JoshuaLampert: I agree that there's no fundamental difference between boxes of different dimensions, but I also don't see a big downside of having one test with a geometry of paramdim > 3 and putting it in the extended testing. So if possible I would keep it just to be sure that everything works dimension-independent.

test/combinations.jl
mikeingold: I just commit a change for all of the `@testitem`s. Since the geometry is stored inside of `testable` already, I made the second arg optional in `runtests(testable::TestableGeometry, supports::SupportStatus = SupportStatus(testable.geometry))' and just dropped the second arg of all calls.

test/combinations.jl
mikeingold: The `@testsnippet Setup` was still used in `test/utils.jl` and `test/floating_point_types.jl`, but after seeing this comment and thinking about it some more, I decided to re-organize a bit. Now we have `@testsnippet`s in each test file where they're required so that all relevant code is in the same source file.

test/combinations.jl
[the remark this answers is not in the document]
mikeingold: I just restored the 4D test.

test/combinations.jl
JoshuaLampert: Sounds good.

JoshuaLampert (approved):
Looks good. Thanks!

test/combinations.jl
JoshuaLampert: Even better. Thanks.

PR #152, opened 2024-12-13
add AD (Enzyme) support via MeshIntegralsEnzymeExt

kylebeggs:
This is a fresh PR continuing the work in #129 because the rebase for that one was driving me insane:

Adds support to calculate the jacobian via Enzyme autodiff #35. This is done by adding a package extension for Enzyme, MeshIntegralsEnzymeExt.


kylebeggs:
I simply created an error for a handful (4) of troublesome geometries with Enzyme so we can at least get this out for the rest of them

JoshuaLampert (changes requested):
Thanks a lot! Please find below some suggestions and comments. The main suggestion is about refactoring the `check_diff_method_support` function.
I would be interested in some benchmarks. Could we maybe (at least temporarily) change the `_default_diff_method` to use `AutoEnzyme` to let the benchmarks script run for the `AutoEnzyme` backend, such that we get a comparison between this and the `FiniteDifference` backend?
This also leads to the question whether we want `AutoEnzyme` to be the `_default_diff_backend` permanently. We could define something like
```julia
function _default_diff_method(g::Type{G}) where {G <: Geometry}
  if supports_autoenzyme(g)
    return AutoEnzyme()
  else
    return FiniteDifference()
  end
end
```
but that's something to discuss and maybe also depends on the benchmark results. (The code snippet above wouldn't run with my suggestions below because they define `supports_autoenzyme(::Geometry)` and not `supports_autoenzyme(::Type{G}) where {G <: Geometry}`, but these are details.)

JoshuaLampert:
You would need to `using Enzyme` in [the benchmark script](https://github.com/JuliaGeometry/MeshIntegrals.jl/blob/c7c0a47fc475b96599c8ed8337f313f4de732830/benchmark/benchmarks.jl) to let the benchmarks load the package extension.

Project.toml
```diff
@@ -12,13 +12,20 @@ Meshes = "eacbb407-ea5a-433e-ab97-5258b1ca43fa"
...
 
+[weakdeps]
+Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
+
+[extensions]
+MeshIntegralsEnzymeExt = "Enzyme"
+
 [compat]
 CliffordNumbers = "0.1.9"
 CoordRefSystems = "0.12, 0.13, 0.14, 0.15, 0.16"
```
JoshuaLampert, line 23:
```suggestion
CoordRefSystems = "0.15, 0.16"
```
This should fix the Downgrade action error.

JoshuaLampert (changes requested):

benchmark/benchmarks.jl
```diff
@@ -3,6 +3,7 @@ using LinearAlgebra
 using Meshes
 using MeshIntegrals
 using Unitful
+using Enzyme
```
JoshuaLampert, line 6:
```suggestion
import Enzyme
```
There is a namespace error with `jacobian`

JoshuaLampert (changes requested):

benchmark/Project.toml
```diff
@@ -1,11 +1,13 @@
...
+Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Meshes = "eacbb407-ea5a-433e-ab97-5258b1ca43fa"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [compat]
 BenchmarkTools = "1.5"
+Enzyme = "0.13"
 LinearAlgebra = "1"
 Meshes = "0.50, 0.51, 0.52"
```
JoshuaLampert, line 12:
```suggestion
Meshes = "0.51.20, 0.52
```

JoshuaLampert (changes requested):

test/Project.toml
```diff
@@ -13,9 +14,10 @@ Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 [compat]
 Aqua = "0.7, 0.8"
 CoordRefSystems = "0.12, 0.13, 0.14, 0.15, 0.16"
```
JoshuaLampert, line 16:
```suggestion
CoordRefSystems = "0.15, 0.16"
```

JoshuaLampert:
The benchmark results look quite nice. Overall mostly a pretty solid speedup :+1:

test/combinations.jl
```diff
@@ -178,6 +184,9 @@ end
     end
     solution = 2π * u"Ω*m"
 
+    # Enzyme support
+    @test MeshIntegrals.supports_autoenzyme(curve) == false
```
JoshuaLampert, line 188: Would these tests maybe be better located in the utils.jl tests?

src/utils.jl
```diff
@@ -12,15 +12,35 @@ end
...
+supports_autoenzyme(::Type{<:Meshes.BezierCurve}) = false
+supports_autoenzyme(::Type{<:Meshes.CylinderSurface}) = false
+supports_autoenzyme(::Type{<:Meshes.Cylinder}) = false
+supports_autoenzyme(::Type{<:Meshes.ParametrizedCurve}) = false
+supports_autoenzyme(::G) where {G <: Geometry} = supports_autoenzyme(G)
+
+_check_diff_method_support(::Geometry, ::DifferentiationMethod) = nothing
+function _check_diff_method_support(geometry::Geometry, ::AutoEnzyme)
+    if !supports_autoenzyme(geometry)
+        throw(ArgumentError("Differentiation method AutoEnzyme not supported for this geometry."))
```
JoshuaLampert, line 25: codecov reports this line is uncovered. Could you maybe add a test for that?

test/combinations.jl
kylebeggs: In my head it makes more sense to group all things relevant to a specific geometry with each other, but if you think we should move it to utils.jl, I will. I don't feel strongly about it!

JoshuaLampert: Yeah, I don't know. I also do not care so much. We can also keep it here if @mikeingold does not object.

Project.toml
```diff
@@ -12,13 +12,20 @@ Meshes = "eacbb407-ea5a-433e-ab97-5258b1ca43fa"
...
+Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
+
+[extensions]
+MeshIntegralsEnzymeExt = "Enzyme"
+
 [compat]
 CliffordNumbers = "0.1.9"
-CoordRefSystems = "0.12, 0.13, 0.14, 0.15, 0.16"
+CoordRefSystems = "0.15, 0.16"
+Enzyme = "0.13"
```
JoshuaLampert, line 24:
```suggestion
Enzyme = "0.13.19"
```
Downgrade still fails. Looks like we need a newer Enzyme version. I just checked locally. I think we need Enzyme v0.13.19.

test/Project.toml
```diff
@@ -12,10 +13,11 @@ Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [compat]
 Aqua = "0.7, 0.8"
-CoordRefSystems = "0.12, 0.13, 0.14, 0.15, 0.16"
+CoordRefSystems = "0.15, 0.16"
+Enzyme = "0.13"
```
JoshuaLampert, line 17:
```suggestion
Enzyme = "0.13.19"
```
See above.

benchmark/Project.toml
```diff
@@ -1,12 +1,14 @@
 [deps]
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
+Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Meshes = "eacbb407-ea5a-433e-ab97-5258b1ca43fa"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [compat]
 BenchmarkTools = "1.5"
+Enzyme = "0.13"
```
JoshuaLampert, line 10:
```suggestion
Enzyme = "0.13.19"
```
See above.

JoshuaLampert (changes requested):

test/combinations.jl
mikeingold: If we're only implementing a smallish number of these tests, then I actually would lean towards moving these tests into `utils.jl` under the `@testitem "Differentiation"`, maybe in their own `@testset` if you think it makes sense. If this property needed to be checked for every geometry type, then maybe we could slide it into the test generation function based on `supports.autoenzyme`.

Prior to the recent refactor of `combinations.jl`, I'd had the same idea to add a few extra tests into these sections wherever they were most convenient, but I found that it made keeping track of what tests were implemented and where more convoluted. In the long-run I think the more organized and structured we can make the tests the better.

Project.toml
JoshuaLampert: I am not sure to be honest. From the [error message](https://github.com/JuliaGeometry/MeshIntegrals.jl/actions/runs/12319562537/job/34386898636#step:8:367 ) it looks like there was a missing method definition, but I wasn't able to find a commit where it was added in Enzyme.jl. What is also a bit strange is that locally when I tested it under Julia v1.11 I had no issues. The issue was only with Julia v1.10 (what we are also using in the Downgrade test). So the error looks a bit more involved to me and is probably not worth spending too much time to understand what exactly happenend.

kylebeggs: Billy Moses is pretty helpful so we probably could get some help from him, but given the rapid pace of development on Enzyme currently, this will be old news soon anyways.

src/utils.jl
kylebeggs: I looked at codecov and it showed a different line? I tried to cover that, lets see how the
report comes out this time.

Project.toml
mikeingold: Weird. Agreed that it’s probably not worth spending too much energy on in the grand scheme, especially for an optional extension and given that the Julia AD ecosystem is pretty fast-moving.

src/differentiation.jl
```diff
@@ -23,12 +23,17 @@ struct FiniteDifference{T <: AbstractFloat} <: DifferentiationMethod
     ε::T
 end
 
-# Default constructors
-FiniteDifference{T}() where {T <: AbstractFloat} = FiniteDifference{T}(T(1e-6))
-FiniteDifference() = FiniteDifference{Float64}()
+# If ε not specified, default to 1e-6
+FiniteDifference() = FiniteDifference(1e-6)
```
mikeingold, line 27: I think this `FiniteDifference` constructor is a carry-over from before the recent commits; see the updated replacement on the left side of the diff.

mikeingold:
I just added some formatting tweaks and docstrings.

Is it possible to make `AutoEnzyme` default based on whether the module is loaded and the geometry/FP are supported?

What currently happens if a user tries to use `AutoEnzyme` without loading the package/extension first?

JoshuaLampert (commented):
I think we are almost done.

src/utils.jl
JoshuaLampert: Everything is covered now :+1:

kylebeggs:
> Is it possible to make AutoEnzyme default based on whether the module is loaded and the geometry/FP are supported?
> 
> What currently happens if a user tries to use AutoEnzyme without loading the package/extension first?

I had this originally, but was fighting with it a bit so I dropped it. If the user tries to use it withoput loading Enzyme, you'll just see a method error, but it won't tell you why that method does not exist. One way around this is to write `jacobian(..., ::AutoEnzyme)` which returns a helpful error message in the main part of the code and then when you load Enzyme, it should overwrite this with the actual jacobian?

kylebeggs:
@mikeingold thanks for the docstrings and sorry about removing the newer `FiniteDifference` constructors.
@JoshuaLampert did we want to add an option to use different modes for Enzyme? I think that could be pretty easy.

kylebeggs:
How come the benchmarks don't print out as a comment here anymore? Does it just take some time?

kylebeggs:
Ok, agreed. I think this may be ready to merge then?

JoshuaLampert (approved):
Thanks a lot @kylebeggs! I would like to see a final approvement by @mikeingold. Then this is good to go!

mikeingold:
> @mikeingold thanks for the docstrings and sorry about removing the newer `FiniteDifference` constructors.

No worries. I was going to add a comment but realized I already had the file open locally, so it was a quick fix.

> @JoshuaLampert did we want to add an option to use different modes for Enzyme? I think that could be pretty easy.

Concur with @JoshuaLampert here. I think this is definitely worth exploring. If it’s a quick change with obvious implementation I’d be up for seeing it here, but it might be worth completing this as-is and then exploring that in a new PR/branch with a tighter focus.

mikeingold (approved):

mikeingold:
Approved. Just let us know what you want to do about the other modes.

PR #157, opened 2024-12-19
Use `hcubature_buffer` for decomposed geometries

mikeingold:
## Changes
- For geometries that must be decomposed into multiple pieces, use `HCubature.hcubature_buffer` to pre-allocate a buffer that can be re-used.

## Conclusion
Performance did not improve when using this for `Rope` or `CylinderSurface`.


mikeingold:
Using this technique seems to be marginally slower for a `Rope`. I'm guessing that the overhead there just isn't worth the savings per-integral.

I'm going to try some other geometries to see if any others get any benefit.

PR #158, opened 2024-12-26
Add a changelog

mikeingold:
### Added

- Implement a `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-No changes yet since v0.16.0.
+### Changed
+
+- Implemented a more efficient internal parametric transformation for `Meshes.Tetrahedron`, resulting in about an 80% integral performance improvement.
 
 
 ## [0.16.0] - 2024-12-14
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ No changes yet since v0.16.0.
   - Reorganized `@testsnippet`s to exist in same source file as relevant tests.
   - Removed `:extended` tag from `Tetrahedron` now that performance is significantly improved.
 
+
 ## [0.15.2] - 2024-10-25
 
 MeshIntegrals.jl is now owned by the JuliaGeometry organization!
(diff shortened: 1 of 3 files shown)


mikeingold:
> I like it! Especially in view of the recent changes to the requirements for registering breaking releases in the General registry. In future, we can, e.g., point to the changelog in the release notes.

This was actually inspired by [the recent thread on Julia Discourse](https://discourse.julialang.org/t/new-automerge-requirement-release-notes-required-for-breaking-package-releases/123955?u=mike.ingold).

> We could go even further and integrate the changelog into the documentation leveraging [Changelog.jl](https://github.com/JuliaDocs/Changelog.jl). This also allows automatic linking to the corresponding PR implementing the respective change. I've done something similar before. If you agree, I could add such a workflow.

I saw that package referenced in the Discourse thread. Is there more to it than just automatically generating links? If not, I don't mind just manually adding links where relevant.

JoshuaLampert:
I think it would be nice to have the changelog also in the docs for more visibility. To achieve this we have different options. We need a markdown file under `docs/src/` for Documenter.jl to pick it up. We could simply duplicate the CHANGELOG.md and copy it in `docs/src/` and every time we add something to the changelog, we also add it in the other file. But this is not so nice because we always have to edit two files. Alternatively, we could programmatically copy the markdown file from the main directory to the docs folder whenever building the docs, i.e. using a `cp` in `make.jl`. This would probably work, but essentially the same thing is done by Changelog.jl by adding something like

```julia
Changelog.generate(Changelog.Documenter(),                    # output type
                   joinpath(@__DIR__, "..", "CHANGELOG.md"),  # input file
                   joinpath(@__DIR__, "src", "changelog.md"); # output file
                   repo = "JuliaGeometry/MeshIntegrals.jl",   # default repository for links
                   branch = "main",)
```

to `make.jl`. Plus it automatically generates the links to the issues/PRs. Thus, the questions are:

1. Do we want to have the changelog in the docs?
2. If yes, which of the three methods above (or another one?) do you prefer?

mikeingold:
I'm not an expert in GitHub and these automations, but could we simply create a symlink like `/docs/src/CHANGELOG.md` pointing to `/CHANGELOG.md`?

JoshuaLampert:
Could maybe work. I've never tried that.

mikeingold:
I can give it a shot here and see what happens.

mikeingold:
Looks like it worked!

mikeingold:
https://juliageometry.github.io/MeshIntegrals.jl/previews/PR158/CHANGELOG/

PR #159, opened 2024-12-27
Use more efficient parametric transformation for `Tetrahedron`

### LINKED ISSUE #145: Possible to speed up `_parametric(::Tetrahedron)`
## Background

I found in #139 that the performance of `_parametric(::Triangle)` improved by an order-of-magnitude by re-formulating the calculation to use fewer calls to Meshes.jl parametric functions. The original version used `t2` to locate points along two edges of the triangle, generated a `Segment` between them, and then returns `segment(t1)` as the desired `Point`. The replacement version simply calculates the equivalent barycentric coordinates `(u, v)` for the same `Point` and then calls `triangle(u, v)`.

The current formulation for `_parametric(::Tetrahedron)` similarly uses `t3` to locate three points on the edges of the tetrahedron to form a triangular `cross_section`, then calls `_parametric(cross_section::Triangle)(t1, t2)` on it.

## Idea

If we can implement something similar for `Tetrahedron` where the barycentric coordinates `(u, v, w)` are calculate directly and `tetrahedron(u, v, w)` is returned, we can expect to see a similarly impressive performance boost for integrals of tetrahedra.

mikeingold:
### Changed

- Implement a new parametric transformation for `Tetrahedron` that first locates the desired point in barycentric space, requiring only one call to the original parametric function.
  - Benchmark performance improved by around 80%.
  - Updated `_parametric(::Triangle)` for consistent styling.
  - Closes #145

### DIFF
--- a/src/specializations/Tetrahedron.jl
+++ b/src/specializations/Tetrahedron.jl
@@ -34,7 +34,16 @@ function _parametric(tetrahedron::Meshes.Tetrahedron)
             throw(DomainError((t₁, t₂, t₃), msg))
         end
 
-        # Take a triangular cross-section at t3
+        """
+        # Algorithm:
+        - Form a barycentric tetrahedron bounded by the points [0, 0, 0], [1, 0, 0],
+            [0, 1, 0], and [0, 0, 1].
+        - Use t₃ to take a triangular cross-section of the tetrahedron at points
+            [t₃, 0, 0], [0, t₃, 0], and [0, 0, t₃].
+        - Use t₂ to take a line segment cross-section of the triangle between
+            points [t₂t₃, 0, t₃ - t₂t₃] and [0, t₂t₃, t₃ - t₂t₃].
+        - Use t₁ to select a point along this line segment, i.e. ā + t₁(b̄ - ā).
+        """
         u₁ = (t₂ * t₃) - (t₁ * t₂ * t₃)
         u₂ = t₁ * t₂ * t₃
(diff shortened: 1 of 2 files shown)


src/specializations/Tetrahedron.jl
```diff
@@ -28,19 +28,26 @@ end
 
 # Map argument domain from [0, 1]³ to Barycentric domain for (::Tetrahedron)(t1, t2, t3)
```
JoshuaLampert, line 29:
```suggestion
# Map argument domain from [0, 1]³ to Barycentric domain for (::Tetrahedron)(t₁, t₂, t₃)
```

src/specializations/Triangle.jl
```diff
@@ -28,14 +28,22 @@ end
 
 # Map argument domain from [0, 1]² to Barycentric domain for (::Triangle)(t1, t2)
```
JoshuaLampert, line 29:
```suggestion
# Map argument domain from [0, 1]² to Barycentric domain for (::Triangle)(t₁, t₂)
```

JoshuaLampert (commented):
Thanks! This is a nice speedup of 80%. Codecov complains about the docstring-like comments. Could we maybe turn those into normal comments `#` to satisfy codecov? Documenter.jl doesn't pick up these docstrings anyway. But if you prefer the status quo I would also be fine.

JoshuaLampert (approved):
LGTM!

mikeingold:
Thanks for the review!

PR #160, opened 2024-12-28
Fix bug where integral fails when Enzyme ext not loaded

mikeingold:
I discovered a bug where a clean install of Meshes without Enzyme will still default to using `AutoEnzyme` but won’t have the appropriate method defined.

I’m currently trying to fix this by signaling whether the extension is loaded via multiple dispatch. I created a `supports_autoenzyme(::Any) = false` method and moved the rest into the extension itself. That way, without the extension loaded, all calls will return false. When the extension is loaded, the methods with more specific argument types will take over.

### DIFF
--- a/test/combinations.jl
+++ b/test/combinations.jl
@@ -48,11 +48,12 @@ This file includes tests for:
     end
 
     # Shortcut constructor for geometries with typical support structure
-    function SupportStatus(
-            g::Geometry;
-            autoenzyme = true # Assume supported unless explicitly flagged
-    )
-        N = Meshes.paramdim(g)
+    function SupportStatus(geometry::G;) where {G <: Geometry}
+        # Check whether AutoEnzyme should be supported, i.e. not on blacklist
+        unsupported_Gs = Union{BezierCurve, Cylinder, CylinderSurface, ParametrizedCurve}
+        autoenzyme = !(G <: unsupported_Gs)
+
+        N = Meshes.paramdim(geometry)
         if N == 1
             # line/curve
             aliases = Bool.((1, 0, 0))
@@ -73,17 +74,13 @@ This file includes tests for:
             aliases = Bool.((0, 0, 0))
             rules = Bool.((0, 1, 1))
             return SupportStatus(aliases..., rules..., autoenzyme)
-        end
-    end
+        end #if
+    end # function
 
     # Generate applicable tests for this geometry
-    function runtests(
-            testable::TestableGeometry;
-            autoenzyme = true, # Assume supported unless explicitly flagged
-            rtol = sqrt(eps())
-    )
+    function runtests(testable::TestableGeometry; rtol = sqrt(eps()))
         # Determine support matrix for this geometry
-        supports = SupportStatus(testable.geometry; autoenzyme = autoenzyme)
+        supports = SupportStatus(testable.geometry)
 
         # Ensure consistency of SupportStatus with supports_autoenzyme
         @test MeshIntegrals.supports_autoenzyme(testable.geometry) == supports.autoenzyme
@@ -344,7 +341,7 @@ end
 
     # Package and run tests
     testable = TestableGeometry(integrand, cyl, solution)
-    runtests(testable; autoenzyme = false)
+    runtests(testable)
 end
 
 @testitem "Meshes.CylinderSurface" setup=[Combinations] begin
@@ -359,7 +356,7 @@ end
 
     # Package and run tests
     testable = TestableGeometry(integrand, cyl, solution)
-    runtests(testable; autoenzyme = false)
+    runtests(testable)
 end
 
 @testitem "Meshes.Disk" setup=[Combinations] begin
@@ -496,9 +493,9 @@ end
 
     # Package and run tests
     testable_cart = TestableGeometry(integrand, curve_cart, solution)
-    runtests(testable_cart; autoenzyme = false)
+    runtests(testable_cart)
     testable_polar = TestableGeometry(integrand, curve_polar, solution)
-    runtests(testable_polar; autoenzyme = false)
+    runtests(testable_polar)
 end
 
 @testitem "Meshes.Plane" setup=[Combinations] begin
--- a/test/combinations.jl
+++ b/test/combinations.jl
@@ -48,7 +48,10 @@ This file includes tests for:
     end
 
     # Shortcut constructor for geometries with typical support structure
-    function SupportStatus(g::Geometry, autoenzyme = MeshIntegrals.supports_autoenzyme(g))
+    function SupportStatus(
+            g::Geometry;
+            autoenzyme = true # Assume supported unless explicitly flagged
+    )
         N = Meshes.paramdim(g)
         if N == 1
             # line/curve
@@ -73,11 +76,15 @@ This file includes tests for:
         end
     end
 
+    # Generate applicable tests for this geometry
     function runtests(
-            testable::TestableGeometry,
-            supports::SupportStatus = SupportStatus(testable.geometry);
+            testable::TestableGeometry;
+            autoenzyme = true, # Assume supported unless explicitly flagged
             rtol = sqrt(eps())
     )
+        # Determine support matrix for this geometry
+        supports = SupportStatus(testable.geometry; autoenzyme = autoenzyme)
+
         # Test alias functions
         for alias in (lineintegral, surfaceintegral, volumeintegral)
             # if supports.alias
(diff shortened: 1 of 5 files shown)


mikeingold:
@JoshuaLampert I'm a little puzzled that the current code here is failing these tests, but I'm not an expert in package extensions. In my mind, the catch-all line

https://github.com/JuliaGeometry/MeshIntegrals.jl/blob/017ecfe440f31b8fb7bf89e63ae03cdf3d6ce89f/src/utils.jl#L25

should always be used when the Enzyme extension isn't loaded, then when loaded the new methods 

https://github.com/JuliaGeometry/MeshIntegrals.jl/blob/017ecfe440f31b8fb7bf89e63ae03cdf3d6ce89f/ext/MeshIntegralsEnzymeExt.jl#L21-L25

should be loaded.

Am I missing something?

test/utils.jl
```diff
@@ -25,12 +25,22 @@ end
...
-    @test _default_diff_method(sphere, BigFloat) isa FiniteDifference
+@testitem "Differentiation (Enzyme extension loaded)" setup=[Utils] begin
+    # _default_diff_method -- using type or instance, Enzyme-supported combination
+    let sphere = Sphere(Point(0, 0, 0), 1.0)
+        @test _default_diff_method(Meshes.Sphere, Float64) isa AutoEnzyme
+        @test _default_diff_method(sphere, Float64) isa AutoEnzyme
+    end
+
+    # _default_diff_method -- Enzyme-unsupported FP types
+    @test _default_diff_method(Meshes.Sphere, Float32) isa FiniteDifference
```
JoshuaLampert, line 36: `Float32` is supported, isn't it?

test/combinations.jl
```diff
@@ -117,17 +117,20 @@ This file includes tests for:
...
+        for (supported, method) in iter_diff_methods
+            f = testable.integrand
+            geometry = testable.geometry
+            sol = testable.solution
+
             if supported
-                @test integral(
-                    testable.integrand, testable.geometry; diff_method = diff_method)≈testable.solution rtol=rtol
+                @test integral(f, geometry; diff_method = method)≈sol rtol=rtol
                 @test MeshIntegrals.supports_autoenzyme(testable.geometry) == true
```
JoshuaLampert, line 131: This should only be tested with the `AutoEnzyme()` backend shouldn't it?

test/utils.jl
mikeingold: Good catch there, that should’ve been `Float16`.

JoshuaLampert:
> @JoshuaLampert I'm a little puzzled that the current code here is failing these tests, but I'm not an expert in package extensions. In my mind, the catch-all line
> 
> https://github.com/JuliaGeometry/MeshIntegrals.jl/blob/017ecfe440f31b8fb7bf89e63ae03cdf3d6ce89f/src/utils.jl#L25
> 
> should always be used when the Enzyme extension isn't loaded, then when loaded the new methods
> 
> https://github.com/JuliaGeometry/MeshIntegrals.jl/blob/017ecfe440f31b8fb7bf89e63ae03cdf3d6ce89f/ext/MeshIntegralsEnzymeExt.jl#L21-L25
> 
> should be loaded.
> 
> Am I missing something?

In principle it works. I just tried:

```julia
julia> using Meshes, MeshIntegrals

julia> MeshIntegrals._default_diff_method(Meshes.Sphere, Float64)
FiniteDifference{Float64}(1.0e-6)

julia> import Enzyme

julia> MeshIntegrals._default_diff_method(Meshes.Sphere, Float64)
AutoEnzyme()
```

but the test doesn't seem to recognize that the extension is *not* loaded. Maybe if we load the extension in another `@testitem`, it is also loaded in another even without using `Enzyme` in the other?

test/combinations.jl
mikeingold: This `@testsnippet` imports Enzyme, so the previous `@test integral` that did not specify a `diff_method` should fall back to the default, which will use `AutoEnzyme()` (except for the few blacklisted cases).

This specific grouping of tests is intended to explicitly iterate through all `diff_method` choices, so I added `FiniteDifference()` to the iters set.

mikeingold: I guess we're actually also missing user-friendly errors for situations where `AutoEnzyme` is specified/defaulted but isn't loaded and supported.

JoshuaLampert: What I mean is that this test will fail (and actually fails in CI) for geometries that do not support Enzyme (e.g., `BezierCurve`), when `method` is `FiniteDifference`.

mikeingold:
Now that I'm seeing the latest CI results...
- for the Enzyme-loaded case it does actually look like the code is behaving as intended but the test is checking for the wrong support state.
- for the Enzyme-not-loaded case, I wonder if the TestItems.jl backend has Enzyme accessible due to the other `@testsnippet`s even if it isn't explicitly "import"ed by the particular `@testitem`.

JoshuaLampert:
> Now that I'm seeing the latest CI results...
> 
>     * for the Enzyme-loaded case it does actually look like the code is behaving as intended but the test is checking for the wrong support state.
> 
>     * for the Enzyme-not-loaded case, I wonder if the TestItems.jl backend has Enzyme accessible due to the other `@testsnippet`s even if it isn't explicitly "import"ed by the particular `@testitem`.

I agree. The first issue should be resolved by the comment I made above. The second is trickier, I think.

test/combinations.jl
mikeingold: I just pushed a commit with changes to the `combinations.jl` tests that I'm hoping will fix this. Instead of relying on the package to tell us the correct support state that we will verify, which could lead to false negatives, I just added a keyword-argument flag triggered in the relevant `@testitem`s.

JoshuaLampert: I don't think this will fix it because above you explicitly set `supported` to always `true` in the `FiniteDifference()` case. The `supports` SupportStatus is not used for the test in this line.

test/combinations.jl
```diff
@@ -117,17 +124,20 @@ This file includes tests for:
...
```
JoshuaLampert, line 138:
```suggestion
	if method isa AutoEnzyme
                @test MeshIntegrals.supports_autoenzyme(testable.geometry) == true
        end
```
Maybe just something like this?

mikeingold: Ooohhhh... yeah, I was just code-blind on this one. The loop is asking questions specifically about Enzyme support, but it's the new `FiniteDifference` iteration that throws an error.

mikeingold: I'm actually just going to drop the `supports_autoenzyme` checks here since those are now checked more thoroughly in `test/utils.jl`.

JoshuaLampert: That's of course also possible.

JoshuaLampert:
Maybe revert [47193a2](https://github.com/JuliaGeometry/MeshIntegrals.jl/pull/160/commits/47193a2e9327ab07ee21e0c95ba6a63dbffb7b26)? I liked the automatically generated `SupportStatus`.

mikeingold:
> Maybe revert [47193a2](https://github.com/JuliaGeometry/MeshIntegrals.jl/pull/160/commits/47193a2e9327ab07ee21e0c95ba6a63dbffb7b26)? I liked the automatically generated `SupportStatus`.

My concern with the prior arrangement is mostly that we're using the codebase itself as the source of truth against which we're attempting to verify behavior. So, if somehow `supports_autoenzyme(::SomeGeometry)` is producing an unintended result, then the behavior of the code changes, but the checks we're running on the code could also change and result in a false negative/pass.

JoshuaLampert:
We could explicitly test `supports_autoenzyme` separately in utils.jl for the different geometries (I actually thought we are doing that already, but it looks like we only test for `_default_diff_method`, which relies on `supports_autoenzyme`). The reason I like the automatically generated `SupportStatus` is that it minimizes the number of places in the tests, where we check for the support status. Ideally, we would only need to change one `false` to `true` in the tests once we implement support for a geometry, which is currently not supported for `AutoEnzyme`. Also, I would argue that all the tests in combinations.jl are more or less to test `integral` and that the correct `integral` implementation is used. I think, all the infrastructure around that and that the differentiation methods are picked up correctly, should be tested in utils.jl.
Having said that this is mostly bikeshedding, and I'm fine either way :smile:

mikeingold:
I restored a `supports_autoenzyme` check because I do think there's value in ensuring that it's tested for all possible geometries.

I'm thinking more about the `SupportStatus` constructor, but my concern more concretely is that the previous form was equivalent to
```julia
supports.autoenzyme = MeshIntegrals.supports_autoenzyme(geometry)
@test MeshIntegrals.supports_autoenzyme(geometry) == supports.autoenzyme
```

Maybe we could, inside the constructor, use something like
```julia
autoenzyme = typeof(geometry) \notin Union{BezierCurve, ...}
```

mikeingold:
I removed the problematic `@testitem` (Enzyme not loaded) and switched to a constructor like I mentioned.

JoshuaLampert (approved):
LGTM

PR #162, opened 2024-12-31
Add analytical unit tests for `Cylinder` and `CylinderSurface`

mikeingold:
### Changed
- Update units tests for `Cylinder` and `CylinderSurface` to include analytical solutions.
- Added `using CoordRefSystems` to `@testsnippet Combinations`.
- Updated `CHANGELOG.md`.

Continues progress on #67


JoshuaLampert (approved):
Looks good (I didn't check the maths, but I assume the solution is correct if tests are successful). Thanks!

PR #165, opened 2025-01-09
Documentation updates

### LINKED ISSUE #23: Need improved documentation

mikeingold:
### Changed
- Replaced the previous Example Usage page with a new Tutorial page.
- Replaced `/docs/src/index.md` with a symlink to `/README.md` for consistency.
- Updated the Support Matrix (relabeled and reordered columns) and notes.
- Added new pages: Integration Rules, Tips.
- Organized the "Developer Notes" Markdown files into a new sub-directory.
- Add BenchmarkTools.jl and `compat` entries to `/docs/Project.toml`

Will close #23

Documenter preview: https://juliageometry.github.io/MeshIntegrals.jl/previews/PR165/


mikeingold:
Work still in progress, but I'm open to feedback on what I've written so far.

docs/src/tutorial.md
```diff
@@ -0,0 +1,109 @@
...
+
+[**MeshIntegrals.jl**](https://github.com/JuliaGeometry/MeshIntegrals.jl) is
+registered in the official Julia General registry, enabling it to be easily
+installed using Julia on an internet-connected computer.
+**MeshIntegrals.jl** should usually be installed in conjunction with
+[**Meshes.jl**](https://github.com/JuliaGeometry/Meshes.jl) and
+[**Unitful.jl**](https://github.com/PainterQubits/Unitful.jl) which provide
+support for geometries and physical units, respectively.
+
+**MeshIntegrals.jl** can be installed while using Julia by entering `pkg` mode
```
JoshuaLampert, line 13:
```suggestion
**MeshIntegrals.jl** can be installed from the Julia REPL by entering `pkg` mode
```
While sounds to me as two processes happen at the same time.

docs/src/tutorial.md
```diff
@@ -0,0 +1,109 @@
...
+
+or by calling the built-in **Pkg.jl** library.
+```julia
+import Pkg
+Pkg.add("MeshIntegrals")
+```
+
+## Basic Usage
+
+Usage of **MeshIntegrals** typically also involves using **Meshes** and **Unitful**,
```
JoshuaLampert, line 27:
```suggestion
Usage of **MeshIntegrals.jl** typically also involves using **Meshes.jl** and **Unitful.jl**,
```
I'd argue the `.jl` is part of the name, so it should always be attached when referring to the package's name.

docs/src/tutorial.md
```diff
@@ -0,0 +1,109 @@
...
+so all three packages will be used in this example.
+
+```julia
```
JoshuaLampert, line 30:
```suggestion
```@example tutorial
```
Maybe this is a good time to switch to an `@example`, such that the code is actually executed and one can see the output in the docs?

docs/src/tutorial.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+### Define a Geometry to Use as an Integration Domain
+
+Here, representing the domain being integrated over, we will define a Bezier
+curve whose path approximates a sine-wave on the xy-plane.
+
+```julia
+N = 361  # number of control points
+xs = range(-π, π, length=N)  # x will be bounded [-π, π] meters
```
JoshuaLampert, line 43:
```suggestion
xs = range(-π, π, length=N)  # x will be bounded in [-π, π] meters
```

docs/src/tutorial.md
```diff
@@ -0,0 +1,109 @@
...
+coordinates are defined in a physical space with meter units and the result is
+a quantity in units of Ohms per meter.
+```math
+f(x, y, z) = \frac{1}{\sqrt{1 + \cos^2(x/\text{m})}} ~ \Omega/\text{m}
+```
+
+Integrand functions are expected to provide a method that takes a single
+`Meshes.Point` argument, so this can be written in Julia as
+
+```julia
```
JoshuaLampert, line 59:
```suggestion
```@example tutorial
```
See above.

docs/src/tutorial.md
```diff
@@ -0,0 +1,109 @@
...
+# Integrand function that outputs in units of Ohms/meter
+function f(p::Meshes.Point)
+    x, y, z = Meshes.to(p)
+    return (1 / sqrt(1 + cos(x / u"m")^2)) * u"Ω/m"
+end
+```
+
+Alternatively, this could be written in the user-friendly notation
+```julia
```
JoshuaLampert, line 68:
```suggestion
```@example tutorial
```

docs/src/tutorial.md
```diff
@@ -0,0 +1,109 @@
...
+f(x, y, z) = (1 / sqrt(1 + cos(x / u"m")^2)) * u"Ω/m"
+f(p::Meshes.Point) = f(Meshes.to(p)...)
+```
+where the required `f(Meshes.Point)` method simply maps to the method `f(x, y, z)`.
+
+### Integrating
+
+This function can be integrated using recommended defaults simply by calling
+```julia
```
JoshuaLampert, line 77:
```suggestion
```@example tutorial
```

docs/src/tutorial.md
```diff
@@ -0,0 +1,109 @@
...
+integral(f, curve)  # -> Approximately 2π Ω
+```
+
+The alias function `lineintegral` works for this geometry since it has one
+parametric dimension. However, the aliases `surfaceintegral` and `volumeintegral`
+will throw an `ArgumentError` since the geometry is not a surface or volume.
+```julia
```
JoshuaLampert, line 84:
```suggestion
```@example tutorial
```

docs/src/tutorial.md
```diff
@@ -0,0 +1,109 @@
...
+surfaceintegral(f, curve)  # -> throws ArgumentError
+
+volumeintegral(f, curve)  # -> throws ArgumentError
+```
+
+An `IntegrationRule` with settings can also be manually specified. The following
+example uses the adaptive Gauss-Kronrod method with a loosened absolute tolerance
+setting of $10^{-4}~\Omega$, which speeds up integration by sacrificing some
+accuracy.
+```julia
```
JoshuaLampert, line 96:
```suggestion
```@example tutorial
```

docs/src/tutorial.md
```diff
@@ -0,0 +1,109 @@
...
+integral(f, curve, GaussKronrod(atol = 1e-4u"Ω")) # -> Approximately (2π ± 1e-4) Ω
+```
+
+The `integral` function and its aliases also support Julia's do-syntax, which
+provides a code block to define a single-use anonymous function and then injects
+it as a first argument to the preceding call. This can be useful if the integrand
+function will not be used outside this integration call.
+```julia
```
JoshuaLampert, line 104:
```suggestion
```@example tutorial
```

JoshuaLampert (changes requested):
This looks already quite nice. Thanks!
I left some small suggestions below.

docs/src/tutorial.md
mikeingold: I was thinking about this but wasn't sure the exact mechanics of `@example`, `jldoctest`, etc and (ironically) I find the documentation for them a bit lacking. Does `@example` print outputs per-block or once for every label at the last usage of that label?

JoshuaLampert: Per block. See, e.g., https://joshualampert.github.io/KernelInterpolation.jl/stable/nodesets/.

mikeingold: Nice example, thanks!

JoshuaLampert:
We also need to add Meshes.jl and Unitful.jl to https://github.com/JuliaGeometry/MeshIntegrals.jl/blob/main/docs/Project.toml.

JoshuaLampert:
Hm, Documenter.jl errors instead of showing the error if there is an error thrown within an `@example` block.

JoshuaLampert:
> Hm, Documenter.jl errors instead of showing the error if there is an error thrown within an @example block.

Maybe we could put these in a separate `julia` block?

mikeingold:
If the labels cross-over between `@example` and `jldoctest` we use that instead, but I don't think printing the entire error message for each call is really desirable here.

mikeingold:
Trying a `jldoctest` now with the same label. Apparently you don't have to print the entire error string; it matches on the beginning of the produced error.

JoshuaLampert:
Looks like the `jldoctest` doesn't recognize the code from the other `@example` blocks even though they have the same label. We could also make everything a `jldocstest`, maybe?

mikeingold:
Since this block is kind of an outlier and fairly simple, I think I'm just going to convert it back to a static `julia`.

docs/src/integration_rules.md
```diff
@@ -0,0 +1,55 @@
...
+```julia
+rule = GaussKronrod()
+```
+
+All standard `QuadGK.quadgk` keyword-argument options are supported. These can be specified when constructing the rule, where the `kwargs` in `GaussKronrod(kwargs...)` is equivalent to `quadgk(f, a, b; kwargs...)`, e.g.:
+```julia
+rule = GaussKronrod(order = 5, rtol = 1e-4)
+```
+
+## [Gauss-Legendre]](@id gausslegendre)
```
JoshuaLampert, line 23:
```suggestion
## [Gauss-Legendre](@id gausslegendre)
```

docs/src/tips.md
```diff
@@ -0,0 +1,27 @@
+# Tips
+
+## Performance
+
+### Use an `atol` setting for integrals with near-zero solutions
+
+By default, `QuadGK.quadgk` uses a relative tolerance (`rtol`) setting to determine when a solution is sufficiently accurate. For integrals whose true solution equal exactly or very nearly zero, the solver can struggle to meet this relative error tolerance and significantly lengthen compute times. In such a case, performance can often be improved by providing an absolute tolerance setting (`atol`).
```
JoshuaLampert, line 7:
```suggestion
By default, `QuadGK.quadgk` uses a relative tolerance (`rtol`) setting to determine when a solution is sufficiently accurate. For integrals whose true solution equal exactly or very nearly zero, the solver can struggle to meet this relative error tolerance and significantly lengthen compute times. In such a case, performance can often be improved by explicitly providing an absolute tolerance setting (`atol`).
```

docs/src/tips.md
```diff
@@ -0,0 +1,39 @@
...
+julia> f(p::Meshes.Point) = p.coords.x * u"N/m"
+f (generic function with 1 method)
+```
+
+Calculating this problem with a default `GaussKronrod()` rule produces a solution that is very close to zero, on the order of $10^{-11}$. However, the result took a non-trivial amount of time due to a large number of memory allocatons for such a trivial problem.
+```julia
+julia> @btime integral(f, segment, GaussKronrod())
+  154.285 ms (28 allocations: 33.70 MiB)
+2.3298940582252377e-11 m N
+```
```
JoshuaLampert, line 32: What might be interesting and worth mentioning is that using AD via Enzyme also speeds up the computation significantly. I believe, this is due to the fact that with AD we are exact and the relative tolerance is reached (much) earlier:

```julia
julia> using Enzyme

julia> @btime integral(f, segment, GaussKronrod())
  1.006 μs (7 allocations: 480 bytes)
0.0 m N
```

mikeingold: This is a really interesting counter-example, actually. In this case it seems like maybe `quadgk` is struggling to meet the required tolerances because the finite-difference `differential` is just noisy enough to be a problem.

I think I'll do some re-wording to move away from the causal connection to exactly-zero-solutions, and just refer to the general phenomena that sometimes it can surprisingly struggle to close on a solution, in which case providing explicit setting tolerances can be beneficial.

mikeingold: Commit. I've been a little pressed for time, but I think the current draft cleans this up a bit.

mikeingold:
I think this basically checks all of the significant goals from #165 (minus a logo, which I'll leave for future work, ideally to something with better artistic skills lol). Let me know how the latest draft sounds.

docs/src/tips.md
```diff
@@ -0,0 +1,63 @@
...
+
+The default/fallback `jacobian` method uses a finite-difference approximation method, which is reasonably performant and compatible with all geometry types. However, this method can potentially introduce a small amount of approximation error into solutions. This method can be explicitly selected via the keyword argument:
+```julia
+integral(f, geometry; diff_method = FiniteDifference())
+```
+
+**MeshIntegrals.jl** includes an extension for [**Enzyme.jl**](https://github.com/EnzymeAD/Enzyme.jl), using it to implement an [automatic differentiation](https://en.wikipedia.org/wiki/Automatic_differentiation)-based `jacobian` method. This method is typically faster and more accurate, but it is not currently compatible with all geometries. This extension will automatically be loaded when **Enzyme.jl** is present in the active Julia environment. When loaded, this method will automatically be used by `integral` whenever it is compatible with the given geometry. This method can be explicitly selected via the keyword argument:
```
JoshuaLampert, line 17:
```suggestion
**MeshIntegrals.jl** includes an extension for [**Enzyme.jl**](https://github.com/EnzymeAD/Enzyme.jl), using it to implement an [automatic differentiation](https://en.wikipedia.org/wiki/Automatic_differentiation)-based `jacobian` method. This method is typically faster and more accurate, but it is not currently compatible with all geometries. This extension will automatically be loaded when **Enzyme.jl** is present in the active Julia environment. When loaded, this method will automatically be used by `integral` whenever it is compatible with the given geometry. This method can also be explicitly selected via the keyword argument:
```
To emphasize again that it is not necessary to explicitly pass the keyword argument?

docs/src/tips.md
```diff
@@ -0,0 +1,63 @@
...
+
+The `GaussKronrod` and `HAdaptiveCuabture` integration rules both make use of adaptive solvers that use relative tolerance (`rtol`) settings to determine when the result is sufficiently precise to return a solution. Under some circumstances these solvers may struggle to converge on a precise enough solution, resulting in extended compute times. This commonly occurs for integrand functions that are numerically unstable or whose solution is zero or near-zero. In such a case, performance can often be improved by explicitly providing an absolute tolerance setting (`atol`).
+
+This can be observed by benchmarking the integration of an integral problem whose true solution is exactly zero.
+```math
+\int_{-1\text{m}}^{1\text{m}} x \left[\tfrac{\text{N}}{\text{m}}\right] ~ \text{d}x
+= 0 \,\text{Nm}
+```
+
+```julia
```
JoshuaLampert, line 39: Maybe we can make this also an `@example`?

JoshuaLampert (commented):
Thanks. I have two small comments.

docs/src/tips.md
mikeingold: I just commit a change to `@repl` blocks to see how those look.

docs/Project.toml
```diff
@@ -1,3 +1,6 @@
 [deps]
+BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+Meshes = "eacbb407-ea5a-433e-ab97-5258b1ca43fa"
 MeshIntegrals = "dadec2fd-bbe0-4da4-9dbe-476c782c8e47"
+Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
```
JoshuaLampert, line 6: Would be good to also have a compat entry in the docs Project.toml for all dependencies (except MeshIntegrals.jl itself).

docs/src/tips.md
JoshuaLampert: Looks good :+1:

docs/Project.toml
```diff
@@ -1,3 +1,12 @@
...
+
+[compat]
+BenchmarkTools = "1"
```
JoshuaLampert, line 9:
```suggestion
BenchmarkTools = "1"
Documenter = "1"
```
I think it's also recommended to have a compat entry for Documenter.jl itself.

JoshuaLampert (approved):
Thanks!

PR #166, opened 2025-01-17
Add more analytical unit tests

mikeingold:
More progress toward #67 
- [x] Ellipsoid
- [x] Hexahedron
- [x] Tetrahedron
- [x] Torus
- [x] Triangle

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,8 +1,8 @@
 [deps]
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
-Meshes = "eacbb407-ea5a-433e-ab97-5258b1ca43fa"
 MeshIntegrals = "dadec2fd-bbe0-4da4-9dbe-476c782c8e47"
+Meshes = "eacbb407-ea5a-433e-ab97-5258b1ca43fa"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [compat]
--- a/test/combinations.jl
+++ b/test/combinations.jl
@@ -687,7 +687,7 @@ end
         x, y, z = ustrip.(u"m", to(p))
         (x + 2y + 3z) * u"A"
     end
-    solution = (1/4) * u"A*m^3"
+    solution = (1//4) * u"A*m^3"
 
     # Package and run tests
     testable = TestableGeometry(integrand, tetrahedron, solution)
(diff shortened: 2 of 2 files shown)


mikeingold:
Not sure why Downgrade fails...

JoshuaLampert:
> Not sure why Downgrade fails...

That's why: https://github.com/JuliaGeometry/MeshIntegrals.jl/actions/runs/12840564494/job/35809354687?pr=166#step:8:477. The necessary diff is also given there. Aqua.jl wants the entries to be ordered alphabetically. This wasn't caught in the previous PR already [because we skip CI for changes under docs/](https://github.com/JuliaGeometry/MeshIntegrals.jl/blob/main/.github/workflows/CI.yml#L13).

JoshuaLampert (commented):
LGTM.
Just out of interest: How did you come up with the functions and their solutions? Did you compute the solutions by hand, did you use some symbolic solver or similar, or did you find them in some reference, or something else?

mikeingold:
> > Not sure why Downgrade fails...
> 
> That's why: https://github.com/JuliaGeometry/MeshIntegrals.jl/actions/runs/12840564494/job/35809354687?pr=166#step:8:477. The necessary diff is also given there. Aqua.jl wants the entries to be ordered alphabetically. This wasn't caught in the previous PR already [because we skip CI for changes under docs/](https://github.com/JuliaGeometry/MeshIntegrals.jl/blob/main/.github/workflows/CI.yml#L13).

Oooohhhhh… I guess they’re using the sorting convention that all upper-case letters come before lower-case letters.

mikeingold:
> Just out of interest: How did you come up with the functions and their solutions? Did you compute the solutions by hand, did you use some symbolic solver or similar, or did you find them in some reference, or something else?

Usually I hand-derive integral problems and just choose an integrand function that produces a non-zero solution but is also still pretty reasonable to derive by hand.

The `Hexahedron` and `Ellipsoid` solutions here came from the observation that both have degenerate cases, i.e. they can just be turned into a cube and a sphere.

The others in this PR actually came from experimenting with an LLM (GPT 4o) which was surprisingly competent at suggesting test cases with step-by-step derived solutions. I just guided the model a bit and then validated the provided solution by hand.

JoshuaLampert:
> > Just out of interest: How did you come up with the functions and their solutions? Did you compute the solutions by hand, did you use some symbolic solver or similar, or did you find them in some reference, or something else?
> 
> Usually I hand-derive integral problems and just choose an integrand function that produces a non-zero solution but is also still pretty reasonable to derive by hand.
> 
> The `Hexahedron` and `Ellipsoid` solutions here came from the observation that both have degenerate cases, i.e. they can just be turned into a cube and a sphere.
> 
> The others in this PR actually came from experimenting with an LLM (GPT 4o) which was surprisingly competent at suggesting test cases with step-by-step derived solutions. I just guided the model a bit and then validated the provided solution by hand.

Sounds good.

JoshuaLampert (approved):

mikeingold:
Thanks!

Issue #169, opened 2025-02-21
Integral on `PolyArea` not yet implemented


mikeingold:
The `Meshes.PolyArea` geometry type doesn't have a parametrization function defined -- and probably won't ever have one. Once `SimpleMesh` integrals are implemented (#27), this could probably also be implemented using something like

```julia
function integral(f, polyarea::PolyArea)
    mesh::SimpleMesh = discretize(polyarea)
    return integral(f, mesh)
end
```

PR #182, opened 2025-06-24
Add support for integrating over `PolyArea` and `Domain` types

### LINKED ISSUE #27: Integral on `SimpleMesh` not yet implemented
Ideas sought. A simple version of this might be possible by simply decomposing the mesh into a set of `Triangle`s and then returning the sum of their integrals, i.e.:

```julia
function _integral_2d(f, mesh::SimpleMesh, rule::IntegrationRule)
    triangle = _some_function_mesh_to_triangles(mesh)
    integrand(triangle) = _integral(f, triangle, rule)
    return sum(integrand, triangles)
end
```

mikeingold:
# Changes

- Adds `integral` methods for `Meshes.PolyArea` and subtypes of `Meshes.Domain` with explicit tests for
  - `CartesianGrid`
  - `PolyArea`
  - `RegularGrid`
  - `SimpleMesh`
- Generalizes the alias functions (e.g. `lineintegral`) to also accept `Domain`s.
- Update docstrings to improve clarity and consistency.

### Closes:
- #27
- #169


mikeingold:
Interesting. The remaining errors with the `SimpleMesh` tests are invalid barycentric coords errors from integrating `Triangle`s.

mikeingold:
Still debugging. One of the remaining issues (`UndefVarError: geometry not defined`) has a stack trace that appeared at first to be throwiing directly from `combinations.jl:138`'s use of `geometry`, but I've done some debugging and it appears to be coming from some other function further along the stack.

mikeingold:
I've run some tests locally to figure out why the `Meshes.CartesianGrid` test set takes so long, and I've run across some odd results.
- The test set initially took about 80 minutes to run on my computer.
- I ran some tests in a separate REPL and couldn't find any obvious slow integration. Then I tried adding some `Logging.@debug` lines to give me a quick trace on where the tests are spending all of their time, and suddenly it would repeatably run in about 15 seconds.
- I manually reverted my local repo state to the current commit to make sure there wasn't some unnoticed change and then I rebooted VS Code. After re-running the test set, now it takes about 12 minutes.

mikeingold:
Based on a little more debugging, even in the longer `CartesianGrid` runs the actual `runtests` function only requires 10-15 seconds, so I suspect that TestItems is also measuring compilation time and that must be dominating the first runtime in this case. Odd that it only seems to be an issue here and not also for `RegularGrid`. Maybe some kind of type instability/inference issue?

mikeingold:
I had a vague suspicion that turned out to be true. The really extreme CI times were nearly 100% compilation time. The output of `elements(discretize(geometry))` is a generator, so I wondered if additionally `collect`ing this array before integration would simplify things for the compiler at the cost of a single small additional allocation. Based on the new CI times this seems to have worked.

mikeingold:
Need to confirm but it seems like the discretization of this test case `PolyArea` includes the hole region... possibly a bug in Meshes?

mikeingold:
Whoa, that was fast lol. No evidence yet but no time to debug at the moment. I constructed a `PolyArea` of a rectangle minus a rectangular hole and the integral value seems to include the outer rectangle plus the inner rectangle (versus the outer minus the inner). Should be easy to debug by just inspection the elements produced.

juliohm:
Make sure that the inner ring is CW and the outer ring is CCW. I believe the PolyArea constructor takes care of this fix, or at least it used to take care. Maybe we removed the fix for maximum performance, don't remember out of my head.

mikeingold:
> Make sure that the inner ring is CW and the outer ring is CCW. I believe the PolyArea constructor takes care of this fix, or at least it used to take care. Maybe we removed the fix for maximum performance, don't remember out of my head.

Interesting. I didn't realize there was a required or preferred orientation for holes. It looks like this was the issue. The construction originally looked like
```julia
@testitem "Meshes.PolyArea" setup=[Combinations] begin
    # Geometry
    a, b, c = 0.4, 0.6, 1
    outer = [(0, 0), (c, 0), (c, c), (0, c)]
    hole = [(a, a), (a, b), (b, b), (b, a)]
    area = PolyArea([outer, hole])
    # ...
end
```
but after changing `hole = [(a, a), (b, a), (b, b), (a, b)]` the correct integral value was returned.

I just ran the previous version (CCW `hole` orientation) in a local Julia REPL and calling `viz(collect(elements(discretize(area))), showsegments=true)` shows overlapping regions with the hole area filled. With the CW `hole` orientation set, `viz` looks as expected with no overlapping regions and an empty center hole.

mikeingold:
@JoshuaLampert: I believe I've gotten past the biggest hurdles on this PR and should have it wrapped up soon. How would you feel about us targeting a new v0.17.0 release that includes this and removal of the obsoleted `integral(f, volume, GaussKronrod())`?

src/specializations/PolyArea.jl
```diff
@@ -0,0 +1,60 @@
+################################################################################
+#                      Specialized Methods for PolyArea
+#
+# Why Specialized?
+#   The PolyArea geometry defines a surface bounded by a polygon but Meshes.jl
+#   can not define a parametric function for a polyarea because a general solution
```
JoshuaLampert, line 6:
```suggestion
#   The PolyArea geometry defines a surface bounded by a polygon, but Meshes.jl
#   cannot define a parametric function for a polyarea because a general solution
```

JoshuaLampert (changes requested):
That's great to hear! Thanks, @mikeingold. This PR is not breaking, is it? So why not release this as v0.16.4 and then we can remove the obsolete `GaussKronrod` for volumes method in v0.17? IMHO, it could be a bit confusing to also add non-breaking changes in a breaking release.

mikeingold:
> This PR is not breaking, is it? So why not release this as v0.16.4 and then we can remove the obsolete `GaussKronrod` for volumes method in v0.17? IMHO, it could be a bit confusing to also add non-breaking changes in a breaking release.

Fair point. I'll plan to release this under `v0.16.4`. If you can think of any breaking changes we should implement with `v0.17.0` in the near future let me know.

juliohm:
It is an internal type for dispatch purposes. Not something intended for
end users.

Em sáb., 19 de jul. de 2025, 14:12, Michael Ingold ***@***.***>
escreveu:

src/integral.jl
```diff
@@ -3,35 +3,50 @@
...
+    integral(f, geometry[, rule]; kwargs...)
 
 Numerically integrate a given function `f(::Point)` over the domain defined by
 a `geometry` using a particular numerical integration `rule` with floating point
 precision of type `FP`.
 
 # Arguments
 - `f`: an integrand function, i.e. any callable with a method `f(::Meshes.Point)`
-- `geometry`: some `Meshes.Geometry` that defines the integration domain
+- `geometry`: a Meshes `Geometry` or `Domain` that defines the integration domain
```
JoshuaLampert, line 14:
```suggestion
- `geometry`: a Meshes.jl `Geometry` or `Domain` that defines the integration domain
```

src/integral.jl
```diff
@@ -3,35 +3,50 @@
...
-use for calculating Jacobians that are used to calculate differential elements
-- `FP = Float64`: the floating point precision desired.
+- `diff_method::DifferentiationMethod`: manually specifies the differentiation method use to
+calculate Jacobians within the integration domain.
+- `FP = Float64`: manually specifies the desired output floating point precision
 """
 function integral end
 
+# Default integration rule to use if unspecified
+_defrule(geometry) = (Meshes.paramdim(geometry) == 1) ? GaussKronrod() : HAdaptiveCubature()
```
JoshuaLampert, line 26:
```suggestion
_default_rule(geometry) = (Meshes.paramdim(geometry) == 1) ? GaussKronrod() : HAdaptiveCubature()
```
For consistency with `_default_diff_method`.

src/integral.jl
```diff
@@ -3,35 +3,50 @@
...
+
 # If only f and geometry are specified, select default rule
 function integral(
         f,
         geometry::Geometry,
-        rule::I = Meshes.paramdim(geometry) == 1 ? GaussKronrod() : HAdaptiveCubature();
+        rule::I = _defrule(geometry);
```
JoshuaLampert, line 32:
```suggestion
        rule::I = _default_rule(geometry);
```

src/integral.jl
```diff
@@ -3,35 +3,50 @@
...
         kwargs...
 ) where {I <: IntegrationRule}
     _integral(f, geometry, rule; kwargs...)
 end
 
+function integral(
+        f,
+        domain::Meshes.Domain,
+        rule::I = _defrule(domain);
```
JoshuaLampert, line 41:
```suggestion
        rule::I = _default_rule(domain);
```

JoshuaLampert (changes requested):
Really nice! I only have some minor comments.

mikeingold:
I accepted the suggestions but had to reformat the `_default_rule` definition to avoid exceeding our max line length formatting rule.

JoshuaLampert (approved):
Thanks!

PR #193, opened 2025-07-30
Improve unit testing for AutoEnzyme support

mikeingold:


mikeingold:
Weird. GitHub CI indicates test failures (no error) on
https://github.com/JuliaGeometry/MeshIntegrals.jl/blob/c04420f6664c340b123548d397ab99f8f4778178/test/utils.jl#L44
and
https://github.com/JuliaGeometry/MeshIntegrals.jl/blob/c04420f6664c340b123548d397ab99f8f4778178/test/utils.jl#L49
but these pass (error) on my local computer.

JoshuaLampert:
Interesting. Do you also use Julia v1.10 and the same versions of Enzyme.jl and Enzyme_jll.jl as in CI?

test/utils.jl
```diff
@@ -26,25 +26,35 @@ end
...
         @test _default_diff_method(Meshes.Sphere, Float64) isa AutoEnzyme
         @test _default_diff_method(sphere, Float64) isa AutoEnzyme
     end
-
-    # _default_diff_method -- Enzyme-unsupported FP types
+    # supports_autoenzyme(::Type{<:Any}) for CI completeness
+    @test MeshIntegrals.supports_autoenzyme(Nothing) == false
+    # BezierCurve: blocked by upstream Meshes-Enzyme incompatibility
+    let curve = BezierCurve([Point(t, 0) for t in range(-π, π, length = 361)])
+        @test_throws "proven readonly" jacobian(curve, (0.5,), AutoEnzyme())
```
wsmoses, line 39: if you know that curve isn't written to, you can say Enzyme.jacobian(Const(curve), ...)

mikeingold: This line uses `MeshIntegrals.jacobian` which maps internally to `Enzyme.jacobian(Enzyme.Forward, curve, ts...)` where `curve` is a callable struct such that `curve(ts...)` yields a point on the curve. Can `EnzymeCore.Const` be applied to `curve` in this context?

wsmoses:
this is a julia version issue, Julia emits different code in the two cases

ext/MeshIntegralsEnzymeExt.jl
mikeingold, line 1: Alphabetized.

wsmoses:
yeah pretty much

test/utils.jl
```diff
@@ -26,25 +26,33 @@ end
...
+    # Cylinder: blocked by version-dependent upstream Meshes-Enzyme incompatibility
+    let cyl = Cylinder(Point(0, 0, 0), Point(0, 0, 1), 2.0)
+        @test _default_diff_method(Meshes.Cylinder, Float64) isa FiniteDifference
+    end
+    # CylinderSurface: blocked by version-dependent upstream Meshes-Enzyme incompatibility
+    let cylsurf = CylinderSurface(Point(0, 0, 0), Point(0, 0, 1), 2.0)
+        @test _default_diff_method(Meshes.CylinderSurface, Float64) isa FiniteDifference
+    end
+    # ParametrizedCurve: no default AD since user-defined functions
+    @test _default_diff_method(Meshes.ParametrizedCurve, Float64) isa FiniteDifference
```
JoshuaLampert, line 51: What would be the error if we want to apply `jacobian` to a `ParametrizedCurve`?

mikeingold: Since these are constructed with user-defined parametric functions, I think the intent here was to use safe defaults since we can’t guarantee Enzyme will work for any arbitrary function, and provide an opt-in via an explicit `diff_method` kwarg.

In hindsight that’s not super consistent with how we treat other geometries whose support can be guaranteed, but I don’t know if it contradicts the docs either.

test/utils.jl
```diff
@@ -25,37 +28,50 @@ end
...
+    end
 
     # Two-argument jacobian
-    segment = Segment(Point(0), Point(1))
-    @test MeshIntegrals.jacobian(segment, (0.5,)) == (Vec(1),)
+    let segment = Segment(Point(0), Point(1))
+        @test MeshIntegrals.jacobian(segment, (0.5,)) == (Vec(1),)
+    end
 
     # Test jacobian with wrong number of parametric coordinates
```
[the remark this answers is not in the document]
JoshuaLampert: That is indeed very scary. According to [this](https://github.com/JuliaGeometry/MeshIntegrals.jl/actions/runs/16842417424/job/47716077762?pr=193#step:2:555) what the bot wants to suggest is what I suggest in the other comment. I have no idea why this suggestion is wrong.

test/utils.jl
```diff
@@ -25,37 +28,50 @@ end
...
-    @test _default_diff_method(Meshes.BezierCurve, Float64) isa FiniteDifference
-    @test _default_diff_method(Meshes.CylinderSurface, Float64) isa FiniteDifference
-    @test _default_diff_method(Meshes.Cylinder, Float64) isa FiniteDifference
-    @test _default_diff_method(Meshes.ParametrizedCurve, Float64) isa FiniteDifference
-
     # FiniteDifference
-    @test FiniteDifference().ε ≈ 1e-6
+    let diff = FiniteDifference()
+        @test typeof(diff.ε) == Float64
+        @test diff.ε  ≈ 1e-6
```
JoshuaLampert, line 62:
```suggestion
        @test diff.ε ≈ 1e-6
```
This is what the formatter suggests.

JoshuaLampert (changes requested):

test/utils.jl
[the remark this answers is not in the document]
mikeingold: Ok, looks like it just got the line number wrong. There were two commits pushed in short succession, so maybe it got confused?

JoshuaLampert (approved):

PR #197, opened 2025-08-14
Update dependency minimum compats

mikeingold:


mikeingold:
Weird. I had to re-run a testing CI the other day because the Coveralls API sent an "under heavy load" response, and now we're getting "internal server error", but

https://status.coveralls.io/

doesn't currently indicate any issues.

JoshuaLampert:
Now, [status.coveralls.io](https://status.coveralls.io/) reports an outage.

mikeingold:
Looks like it's back up and functional again.

